Issue #2, opened 2021-09-26
Make CW721 equivalent to ERC721-Metadata


iLoveBusinessDevelopment:
The current implementation of CW721 is not directly compatible with ERC721 with metadata. The proposal is to remove some fields and add a `token_uri` field for a metadata extension.

Collection-wide fields for CW721:
- name (name of the NFT collection)
- symbol (symbol of NFT collection)

For each `TokenId` expect it to point to one field `token_uri`, like [ERC721](https://docs.openzeppelin.com/contracts/4.x/api/token/erc721#IERC721Metadata)

So mint would look like this:
```
#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
pub struct MintMsg {
    /// Unique ID of the NFT
    pub token_id: String,
    /// The owner of the newly minter NFT
    pub owner: String,
    /// A distinct Uniform Resource Identifier (URI) for a given asset.
    pub token_uri: String,
}
```

We expect the `token_uri` to point to a JSON file that conforms to the CW721 JSON schema (a direct copy of the ERC721 JSON schema):
```
{
    "title": "Asset Metadata",
    "type": "object",
    "properties": {
        "name": {
            "type": "string",
            "description": "Identifies the asset to which this NFT represents"
        },
        "description": {
            "type": "string",
            "description": "Describes the asset to which this NFT represents"
        },
        "image": {
            "type": "string",
            "description": "A URI pointing to a resource with mime type image/* representing the asset to which this NFT represents. Consider making any images at a width between 320 and 1080 pixels and aspect ratio between 1.91:1 and 4:5 inclusive."
        }
    }
}
```
This where the fields `name`, `description`, and `image` go. Individual NFT creators can then add new JSON fields to extend their NFT:

For example adding an `attributes` field to support traits:

```
{
    "name": "name-of-my-asset",
    "description": "description-of-my-asset",
    "image": "image-url",
    "attributes": [
        {
          "trait_type": "Name of some trait", 
          "value": "Value of some trait"
        }, 
    ]
}
```
In general, extensions to the CW721 JSON schema should follow the [opensea metadata standard](https://docs.opensea.io/docs/metadata-standards).

[...]

PR #3, opened 2021-09-28
Add token_uri, on-chain metadata example

iLoveBusinessDevelopment:
Implements the changes described here: https://github.com/CosmWasm/cw-nfts/issues/2

`cw721_base` -- removes `name`, `description`, and `image`, only top level field is `token_uri` and `extension`. 
`cw721_metadata_onchain` -- example of Extension that moves metadata on-chain, supports name, description, image, and more


the-frey:
At a skim, LGTM. Think the CircleCI config still needs updating however as `ci/circleci: contract_cw721_metadata_uri` has been removed AFAICT

JakeHartnell:
Is there are reason the permissions on all the files changed?

JakeHartnell:
Aside from the question about file permissions, this looks good to me.

JakeHartnell:
Would be nice to validate the `token_uri` is indeed a valid URI and not a random string... but we can do this in a separate ticket if you wish.

iLoveBusinessDevelopment:
> Is there are reason the permissions on all the files changed?

Oops, permission changes reverted.

> Would be nice to validate the `token_uri` is indeed a valid URI and not a random string... but we can do this in a separate ticket if you wish.

`token_uri` can be interpreted arbitrarily. Maybe you would try to check for http:// and https://, but now many NFTs use ipfs://. Something new can come out. Best just leave validating this up to the off-chain querier, who can change their validation logic to fit with the times.

JakeHartnell (approved):
LGTM. Thanks @alwin-peng!

the-frey:
Agree with that. URIs are difficult to have a strong opinion on WRT validity and it's probably sensible to push that onto clients

iLoveBusinessDevelopment:
gonna close and reopen to see if it fixes CI

ethanfrey:
Just saw the CI issue... I had configured status checks in github. Fixed now.

I didn't get a chance to review today, but at least I could make your CI green

the-frey:
What's left now before we merge this? Just Ethan's sense check?

contracts/cw721-metadata-onchain/src/lib.rs
```diff
@@ -4,11 +4,29 @@ use serde::{Deserialize, Serialize};
...
+    pub external_url: Option<String>,
+    pub description: Option<String>,
+    pub name: Option<String>,
+    pub attributes: Option<Vec<Trait>>,
+    pub background_color: Option<String>,
+    pub animation_url: Option<String>,
+    pub youtube_url: Option<String>,
+}
+
+pub type Extension = Option<Metadata>;
```
ethanfrey, line 28: The only reason to use this over cw721-base is to use Metadata. I would make it required. (not an Option)

contracts/cw721-metadata-onchain/src/lib.rs
```diff
@@ -4,11 +4,29 @@ use serde::{Deserialize, Serialize};
...
-    pub metadata_uri: String,
+#[derive(Serialize, Deserialize, Clone, PartialEq, JsonSchema, Debug)]
+pub struct Trait {
+    pub display_type: Option<String>,
+    pub trait_type: String,
+    pub value: String,
 }
 
+// see: https://docs.opensea.io/docs/metadata-standards
+#[derive(Serialize, Deserialize, Clone, PartialEq, JsonSchema, Debug)]
```
ethanfrey, line 15: As an example, this makes sense, that we store it in the exact same format that OpenSea has it.

However, if we build more complex NFTs, we would want to keep this `Metadata` struct as the query result from NFT Info (or have a custom query that just returns it). And maybe store it in a different format, that is more tied to the business case and just transform it to this format for display.

I am thinking Royalties. It is possible to store them in a Trait, but sub-optimal.
However, they could easily be converted to a Trait when serving a query

contracts/cw721-metadata-onchain/src/lib.rs
```diff
@@ -76,22 +94,26 @@ mod tests {
...
             owner: "john".to_string(),
-            name: "Starship USS Enterprise".to_string(),
-            description: Some("Spaceship with Warp Drive".into()),
-            image: None,
-            extension: Extension {
-                metadata_uri: "http://starships.example.com/Starship/Enterprise.json".into(),
-            },
+            token_uri: Some("https://starships.example.com/Starship/Enterprise.json".into()),
+            extension: Some(Metadata {
+                image: None,
```
ethanfrey, line 99: This is a lot of `None`s... to save your fingers, I would:

1. `#[derive(Default)]` for Metadata and Trait
2. Auto-fill default fields:

```rust
Some(Metadata{
    description: Some("Spaceship with Warp Drive".into()),
    name: Some("Starship USS Enterprise".to_string()),
    ..Metadata::default()
})
```

ethanfrey: This doesn't change the function at all, but is nice to read and write

ethanfrey (approved):
Fine with me.

Added a few comments on code style you may want to integrate. But nothing blocking a merge

contracts/cw721-metadata-onchain/src/lib.rs
the-frey: I suppose there's also the question, since these are all Options anyway, whether we include the royalty fields for 2981 by default, and then simply leave it up to the implementer to define the custom queries that make up the rest of that royalty interface... 🤔

contracts/cw721-metadata-onchain/src/lib.rs
iLoveBusinessDevelopment: main reason is to interoperate with cw721-base better, so people can use `NftInfoResponse<Extension>` from `cw721-metadata-onchain` to query both cw721-base and cw721-metadata-onchain

contracts/cw721-metadata-onchain/src/lib.rs
iLoveBusinessDevelopment: will do in a later PR..

PR #8, opened 2021-10-01
Update CW721 base helpers file

findolor:
Updated the helpers file based on https://github.com/CosmWasm/cw-plus/blob/main/contracts/cw20-base/helpers.ts


contracts/cw721-base/helpers.ts
```diff
@@ -1,107 +1,100 @@
...
 interface Network {
-  setup: (password: string, filename?: string) => Promise<SigningCosmWasmClient>
+  setup: (password: string, filename?: string) => Promise<[string, SigningCosmWasmClient]>
   recoverMnemonic: (password: string, filename?: string) => Promise<string>
 }
 
 const useOptions = (options: Options): Network => {
 
-  const loadOrCreateWallet = async (options: Options, filename: string, password: string): Promise<Secp256k1HdWallet> => {
+  const loadOrCreateWallet = async (options: Options, filename: string, password: string): Promise<DirectSecp256k1HdWallet> => {
```
JakeHartnell, line 74: IMO this helper file is doing too much. Would rather have the helpers take signing client and then get rid of all this stuff related to wallet creation, faucets, and config.

There is already the `@cosmicdapps/logic` package which handles all that. Would rather these helpers _only_ handle interacting with the chain when given a `CosmWasmClient` or `SigningCosmWasm` client.

Maybe delete all this stuff? Not sure if others have strong opinions here.

On a side note, been brain storming with `ebaker` about how to auto-generate this stuff from the contract schema.

orkunkl: I agree with you. We are doing a lot of copy paste. It would be better to collect boilerplate somehwere.

the-frey: maybe a base helper in `cw-plus` if it needs collecting somewhere for the purposes of documentation or tutorials?

the-frey:
Purely from a code perspective, the rest of the helpers look okay, with the caveat that I've not had the time to run it on pebblenet or juno and confirm it definitely works.

contracts/cw721-base/helpers.ts
orkunkl: `base-helper.ts` I like this idea

orkunkl:
Let's merge the PR now, and move the base-helpers.ts later

orkunkl (dismissed):
LGTM

JakeHartnell (dismissed):
Do want to make base helpers, but this is fine for now.

the-frey (dismissed):
Agreed, `base-helpers` can be a separate ticket/PR

orkunkl (approved):

PR #12, opened 2021-10-11
feat: added prev_owner to response log

### LINKED ISSUE #11: can add prev_owner/from attributes at transfer_nft and send_nft?
When owner approve the token and contract execute transfer, we can't know where is the token came from by the logs.

This is really hard to track token transfer. I made an example on terra.

1. User A make approve to transfer the cw721 token to contract B  
https://finder.terra.money/columbus-5/tx/14D4251F41053E7572A740D7FDB1760A2D3FD593870B0D70FB1EDBCA7121F211

2. User C execute the contract B to send it to B
https://finder.terra.money/columbus-5/tx/AEA9F3EEF1AD4A29644355777BD4157422E55E50130C2D330BC6EC0F4E63F51F

The token transferred to B from A. But there is no information about that at tx of 2.

Because of this Cw20 have transfer_from and send_from.

It is possible to make the attribute that can show the previous owner?

edison0xyz:
Resolves https://github.com/CosmWasm/cw-nfts/issues/11

Useful feature and have added it in.

### DIFF
--- a/contracts/cw721-base/src/contract_tests.rs
+++ b/contracts/cw721-base/src/contract_tests.rs
@@ -194,6 +194,7 @@ fn transferring_nft() {
         Response::new()
             .add_attribute("action", "transfer_nft")
             .add_attribute("sender", "venus")
+            .add_attribute("from", "venus")
             .add_attribute("recipient", "random")
             .add_attribute("token_id", token_id)
     );
--- a/contracts/cw721-base/src/execute.rs
+++ b/contracts/cw721-base/src/execute.rs
@@ -128,11 +128,15 @@ where
         recipient: String,
         token_id: String,
     ) -> Result<Response<C>, ContractError> {
+        let token_info = self.tokens.load(deps.storage, &token_id)?;
+        let owner = token_info.owner.to_string();
+
         self._transfer_nft(deps, &env, &info, &recipient, &token_id)?;
 
         Ok(Response::new()
             .add_attribute("action", "transfer_nft")
             .add_attribute("sender", info.sender)
+            .add_attribute("from", owner)
             .add_attribute("recipient", recipient)
             .add_attribute("token_id", token_id))
     }


contracts/cw721-base/src/contract_tests.rs
```diff
@@ -194,6 +194,7 @@ fn transferring_nft() {
         Response::new()
             .add_attribute("action", "transfer_nft")
             .add_attribute("sender", "venus")
```
JakeHartnell, line 196: Language is a bit confusing her now... we have `sender`, `ricpient`, and `from`...

JakeHartnell, line 196: Almost like @larry0x's comment of using `from` and `to` like the Ethereum transfer events.

It is important to maintain sender, as that may be different than the owner, even though there is a little duplication there.

JakeHartnell: Also, whatever we decide should be consistent with `SendNft`.

JakeHartnell (changes requested):
`TransferNFT` needs to be consistent with `SendNFT`.

JakeHartnell:
What do folks think about using `to` over `recipient`? Mirroring Ethereum: https://ethereum.org/en/developers/docs/standards/tokens/erc-721/#events

the-frey:
FWIW I'm in favour of sticking close to the ETH standards where possible. `from` and `to` are probably the way to go in that case IIUC.

iLoveBusinessDevelopment:
I think we should continue with `recipient`. CW20 is already widely used at this point and uses `recipient`.

the-frey:
Hmmm, I suppose I'm looking at the POV of devs who are coming from ETH/ERC721 land. Although can see the argument for CW20 style too 🤷

edison0xyz:
I think that we might be better to stick with the current CW standards.
Though i see the point of standardising with the Ethereum contracts, but there should be consistency amongst the CosmWasm contracts.

Changing everything to Ethereum sounds like a lot of work in code-change and communication with exisiting protocols. So perhaps let's avoid that.

orkunkl (approved):
LGTM

ALPAC-4:
And I think we need to add owner at mint response log

https://github.com/CosmWasm/cw-nfts/blob/2712c05fbc365a44809c904f8508410c4f70383e/contracts/cw721-base/src/execute.rs#L111

I think owner is more important information than minter. Minter is always same for same contract

shanev:
This issue is a bit stale. Okay to close?

JakeHartnell:
> This issue is a bit stale. Okay to close?

Yeah.

peara:
@JakeHartnell @larry0x  @shanev Please reconsider this PR. This is very useful for indexers and apps and for the consistency of cw721 as well.
I think adding `owner` as already implemented is good enough. The only issue remaining is adding the same to `send_nft` and `mint`, is that right?

PR #23, opened 2021-11-04
add burn

iLoveBusinessDevelopment:


JakeHartnell (approved):
LGTM

the-frey (approved):
From my POV the code here is fine, and I will go with the mood of the room on the following question:

Does burn belong in base, or should it be an extension?

the-frey:
I know @JakeHartnell is likely to argue for base, but FWIW it isn't in the original EIP-721 and I think likely the majority of contracts would like the ability to burn to be absent, if anything.

Issue #7, opened 2021-10-01
CW-2981


the-frey:
Looking at porting this work on equivalence with EIP-2981 emerging royalty standard to this new repo.

https://github.com/CosmWasm/cw-plus/pull/414

However there are some additional things to consider (the query handlers part is, I think, uncontroversial) WRT the metadata that might sit at token level

From discord:

> given that everything is an Option I wonder if rather than implementing royalty fields separately it should be a part of those onchain meta fields? (see #6)

Essentially we've now added a set of baseline metadata fields to keep in line with Opensea - my question is, that as these are all `Option` typed anyway, whether royalties should be in this default meta -

> looking at porting the royalty stuff I did for cw-plus to the new repo and it seems like cos it's optional maybe we should just be encouraging royalty data (even if None) to be a part of the interface for consideration
> consideration by implementers that is

Don't necessarily have a strong opinion here, but we have an opportunity to set the expectation that marketplace implementers _should be checking_ for royalty info by default on any sales in the Cosmos ecosystem and that maybe is something that is valuable from the perspective of creators.

Obviously for NFT types that royalties don't apply to, it's trivial to just use base, or exclude from the meta.

The big question is probably one at contract level - since that's where you signal whether royalties should be checked (if I remember/understand the EIP correctly https://eips.ethereum.org/EIPS/eip-2981) but this can easily be left as either

a) a hook for contract implementer to optionally implement
b) a default implementation that returns `false` and can be imported
c) a requirement for implementing for any contract that uses meta (i.e. up to the implementer whether it's hard coded, passed in instantiation, etc)

In the pull referenced above,

https://github.com/CosmWasm/cw-plus/blob/d7041a755bf0082744af684f169f67256b024ed0/contracts/cw2981-base/src/msg.rs#L58

is implemented 

https://github.com/CosmWasm/cw-plus/blob/d7041a755bf0082744af684f169f67256b024ed0/contracts/cw2981-base/src/contract.rs#L227

Because in the case of a base implementation you can expect that if royalties are _at token level_ rather than a flat, contract-wide royalty, then the correct thing to do is _always check_. However, of course, this is an example implementation.

the-frey:
thoughts? @JakeHartnell / @orkunkl / @alwin-peng

orkunkl:
I don't think including we need this feature in cw721-base. It is easy to determine if a smart contract supports royalty, send `CheckRoyalty` query msg, if you get a true response great, if error it means not implemented. 
[...]

PR #24, opened 2021-11-06
CW-2981 extension

the-frey:
Appreciate this might need some refining and/or further discussion.

A minimal example of the [EIP-2981 spec](https://eips.ethereum.org/EIPS/eip-2981) implemented in CW-land. i.e. #7 

It's a very light wrapper/extension for CW721 compliant contracts that adds two additional query messages:
- `RoyaltyInfo` to be called when selling an NFT, to find out what royalties are owed
- `CheckRoyalties` to be called on the contract, to see if it implements CW-2981

This adds royalty information at token mint time.

The main thing are the queries that are called when the marketplace sells an NFT. Different contracts might implement them differently, e.g. to do more complex logic than the straight percentage here, but the returned queries should conform to the shape here.

Aside - could move those (and the extension handlers) to a package, if it makes them more reusable.


contracts/cw2981-royalties/src/lib.rs
```diff
@@ -0,0 +1,260 @@
...
+    pub description: Option<String>,
+    pub name: Option<String>,
+    pub attributes: Option<Vec<Trait>>,
+    pub background_color: Option<String>,
+    pub animation_url: Option<String>,
+    pub youtube_url: Option<String>,
+    /// specify whether royalties are set on this token
+    pub royalty_payments: bool,
+    /// This is how much the minter takes as a cut when sold
+    pub royalty_percentage: Option<u32>,
```
JakeHartnell, line 36: Maybe make a `Decimal` type?

contracts/cw2981-royalties/src/lib.rs
```diff
@@ -0,0 +1,260 @@
+pub mod msg;
+
+use crate::msg::{CheckRoyaltiesResponse, RoyaltiesInfoResponse};
+use cosmwasm_std::{to_binary, Deps, StdResult};
+use percentage::Percentage;
```
JakeHartnell: Use `cosmwasm_std::{Decimal}` for one less dependency.

contracts/cw2981-royalties/src/msg.rs
```diff
@@ -0,0 +1,138 @@
...
+                limit,
+            } => CW721QueryMsg::Tokens {
+                owner,
+                start_after,
+                limit,
+            },
+            Cw2981QueryMsg::AllTokens { start_after, limit } => {
+                CW721QueryMsg::AllTokens { start_after, limit }
+            }
+            _ => panic!("cannot covert {:?} to CW721QueryMsg", msg),
```
JakeHartnell, line 123: Don't think we want a panic here...

JakeHartnell (commented):
Small comment about types, but otherwise LGTM.

contracts/cw2981-royalties/src/lib.rs
```diff
@@ -0,0 +1,260 @@
...
+
+    Ok(RoyaltiesInfoResponse {
+        address: royalty_address,
+        royalty_amount: royalty_from_sale_price,
+    })
+}
+
+pub fn check_royalties(_deps: Deps) -> StdResult<CheckRoyaltiesResponse> {
+    Ok(CheckRoyaltiesResponse {
+        royalty_payments: true,
```
JakeHartnell: Is this actually checking anything?

contracts/cw2981-royalties/src/msg.rs
```diff
@@ -0,0 +1,138 @@
...
+    // Should be called on sale to see if royalties are owed
+    // by the marketplace selling the NFT.
+    // See https://eips.ethereum.org/EIPS/eip-2981
+    RoyaltyInfo {
+        token_id: String,
+        // the denom of this sale must also be the denom returned by RoyaltiesInfoResponse
+        // this was originally implemented as a Coin
+        // however that would mean you couldn't buy using CW20s
+        // as CW20 is just mapping of addr -> balance
+        sale_price: u128,
```
JakeHartnell: Should this be a `Uint128`? Might be nicer.

contracts/cw2981-royalties/src/msg.rs
```diff
@@ -0,0 +1,138 @@
...
+        }
+    }
+}
+
+#[derive(Serialize, Deserialize, Clone, PartialEq, JsonSchema, Debug)]
+pub struct RoyaltiesInfoResponse {
+    pub address: String,
+    // Note that this must be the same denom as that passed in to RoyaltyInfo
+    // rounding up or down is at the discretion of the implementer
+    pub royalty_amount: u128,
```
JakeHartnell: Use cosmwasm std `Unit128`?

contracts/cw2981-royalties/src/msg.rs
the-frey: Yep, good point

contracts/cw2981-royalties/src/lib.rs
the-frey: For this contract, as I reasoned it there's nothing to check. This hook is expected to be present to check if the contract does implement CW2981 and signal that on sale royalties should be checked. I think with the implementation at token level it should always return `true` because it's up to the token.

Of course contracts that extend this can determine their own behaviour and replace this function if they have more complex behaviour (for example, you could maintain a secondary index of which tokens actually have royalties, or something).

In this super simple case that isn't necessary however.

This does make me think this explanation should be in the readme, however :joy: as well as perhaps the point about the royalty addr implicitly being able to be an addr/dao/multisig

the-frey:
Will address the comments and add some more tests 👍

contracts/cw2981-royalties/src/lib.rs
JakeHartnell: Hmmmmm, interesting. Yeah maybe add a note. :joy:

contracts/cw2981-royalties/src/msg.rs
the-frey: You need it to match all paths, even though it should never be hit because you extend the outer enum.

JakeHartnell (dismissed):

the-frey:
If this is GTM could it be merged? Otherwise before long I'll have to update the PR because things will have shifted.... 🙃

JakeHartnell:
> If this is GTM could it be merged? Otherwise before long I'll have to update the PR because things will have shifted.... upside_down_face

Well if things shift it will need to be updated anyway. :upside_down_face:

Needs one more thumbs up.

contracts/cw2981-royalties/src/lib.rs
```diff
@@ -0,0 +1,255 @@
...
+    pub description: Option<String>,
+    pub name: Option<String>,
+    pub attributes: Option<Vec<Trait>>,
+    pub background_color: Option<String>,
+    pub animation_url: Option<String>,
+    pub youtube_url: Option<String>,
+    /// specify whether royalties are set on this token
+    pub royalty_payments: bool,
+    /// This is how much the minter takes as a cut when sold
+    pub royalty_percentage: Option<u64>,
```
ethanfrey, line 36: Question: Why do we need the `bool` above if this is an option?

I figure `royalty_payments = royalty_percentage.is_some()`
Also, only adding optional fields makes it more compatible with other contracts (they just happen to have it unset)

contracts/cw2981-royalties/src/msg.rs
```diff
@@ -0,0 +1,139 @@
+use cosmwasm_std::Uint128;
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+use cw721_base::msg::QueryMsg as CW721QueryMsg;
+
+#[derive(Serialize, Deserialize, Clone, PartialEq, JsonSchema, Debug)]
+#[serde(rename_all = "snake_case")]
+pub enum Cw2981QueryMsg {
```
ethanfrey, line 9: I wish there were a better way to extend enums.
But, alas, there currently is not.

contracts/cw2981-royalties/src/query.rs
```diff
@@ -0,0 +1,42 @@
...
+
+    Ok(RoyaltiesInfoResponse {
+        address: royalty_address,
+        royalty_amount: royalty_from_sale_price,
+    })
+}
+
+pub fn check_royalties(_deps: Deps) -> StdResult<CheckRoyaltiesResponse> {
+    Ok(CheckRoyaltiesResponse {
+        royalty_payments: true,
```
ethanfrey, line 45: Wait, always true?

What about `ext.royalty_percentage.is_some()`

ethanfrey (dismissed):
A few points that could use cleanup.
Otherwise fine.

contracts/cw2981-royalties/src/msg.rs
shanev: Haven't tested this with a production build yet, but you can extend an enum using serde `untagged` with an extra variant for the extended functionality. So this has the same effect as `flatten`. 

```rs
#[serde(untagged)]
pub enum QueryMsg {
    Base(cw721_base::QueryMsg),
    Extended(ExtendedQueryMsg),
}
```

ethanfrey: Please try that was `cargo wasm` and let me know.
I bet it hits a limit of serde-json-wasm, but if not, that is a great work-around

ben2x4: I think this may not work as expected as it will only affect the serialization of the enum to JSON. However, when dealing with a contract calling another contract, I believe you will still need the "tagged" compound enum. Need to test this, but I think if a second contract tried to interact with this contract believing it was an unextended cw721_base it would fail unless for some reason cosmwasm leverages Serde for internal calls. 

TLDR this might work from a client's perspective, but I don't think it would work from a second contract's perspective. Someone should really write a macro for this!

contracts/cw2981-royalties/src/lib.rs
the-frey: Good point. I guess as long as the query handler returns the result of that lookup, then the interface remains the same. Will change.

contracts/cw2981-royalties/src/msg.rs
the-frey: I will give it a try and see if the workaround works 🤔

contracts/cw2981-royalties/src/query.rs
the-frey: Me being a little lazy as it's an example contract - it's a check at contract level, for this one the answer to "should I see if, when a token controlled by this contract is sold, royalties are owed?" will always be yes. I think `ext.royalty_percentage.is_some()` would be relevant if you were querying on the `token_id`. Then again, this is based on my interpretation of the granularity of the EIP spec, so I could have it wrong 🙂

contracts/cw2981-royalties/src/msg.rs
JakeHartnell: >  However, when dealing with a contract calling another contract, I believe you will still need the "tagged" compound enum.

This is correct. :cry:

Yeah, there really needs to be a macro for this.

contracts/cw2981-royalties/src/query.rs
JakeHartnell: Seems like this is a case where we don't need to follow Ethereum exactly. The boolean seems redundant.

the-frey: No, it will still return boolean regardless - that's the interface, I'm just saying that `ext.royalty_percentage.is_some()` is at a different level of granularity - token, vs contract

contracts/cw2981-royalties/src/msg.rs
the-frey: Ah yeah, that makes sense now I've taken the time to actually parse what @ben2x4 was getting at 😅

contracts/cw2981-royalties/src/lib.rs
```diff
@@ -0,0 +1,255 @@
...
+    pub fn query(deps: Deps, env: Env, msg: Cw2981QueryMsg) -> StdResult<Binary> {
+        match msg {
+            Cw2981QueryMsg::RoyaltyInfo {
+                token_id,
+                sale_price,
+            } => to_binary(&query_royalties_info(deps, token_id, sale_price)?),
+            Cw2981QueryMsg::CheckRoyalties {} => to_binary(&check_royalties(deps)?),
+            _ => Cw2981Contract::default()
+                .query(deps, env, msg.into())
+                .map_err(|err| err),
```
shanev: Is this necessary? `query` should return the expected result type.

contracts/cw2981-royalties/src/msg.rs
shanev: @ethanfrey It didn't work. Seemed too good to be true :) Will have to dive into serde-json-wasm one of these days and figure it out.

@ben2x4 The tagged unextended and untagged extended should be represented exactly the same during transport and survive serialization and deserialization. That was the hope anyway. Have to dig into the guts of the expansive Wasmverse.

contracts/cw2981-royalties/src/msg.rs
```diff
@@ -0,0 +1,139 @@
...
+                limit,
+            } => CW721QueryMsg::Tokens {
+                owner,
+                start_after,
+                limit,
+            },
+            Cw2981QueryMsg::AllTokens { start_after, limit } => {
+                CW721QueryMsg::AllTokens { start_after, limit }
+            }
+            _ => panic!("cannot covert {:?} to CW721QueryMsg", msg),
```
shanev, line 121: `_ => unreachable!()` would give the compiler a bit more info.

contracts/cw2981-royalties/src/lib.rs
the-frey: you mean `map_err`? or the `into`?

the-frey: Yeah seems like you're right, 'map_err` not needed

the-frey:
Okay I think that's all the review points closed off.

contracts/cw2981-royalties/src/query.rs
the-frey: Rechecked the spec and this interface needs to return a boolean, so yeah the answer to:

>  Checking if the NFT being sold on your marketplace implemented royalties 

For a function that takes a contract and returns a boolean - this _is_ the correct base case, and then if contracts want to implement custom logic, they can. This is just the most basic correct example, which is that when token is sold, it signals you should call `query_royalties_info` with the token ID.

Happy to write a better description of this in the README or what have you, but the important thing is to set the expectation from marketplaces that they should check, _in case there is more complex behaviour than a default `true`_.

In some ways, this repo is also a way of documenting How To Do Things With NFTs:tm: so I'm not that worried that the base case is ostensibly redundant, just that it is enumerated for those that have a requirement for more custom logic in future.

To that end, I will add a better comment on it in the code now.

orkunkl:
@the-frey Do you think we can get this PR in this year xp

the-frey:
@orkunkl I mean, if people want to review it then maybe we can get it in lol

orkunkl (dismissed):
Now it's time...

the-frey:
Okay @orkunkl and @JakeHartnell have updated again with changes from main which means it has cleared your reviews. If you are both still ok to merge, could you review pls 🙏

contracts/cw2981-royalties/src/query.rs
JakeHartnell: Still not a big fan of the `CheckRoyalties` query as I think it's mostly pointless. It's more important in the context of Ethereum.

Could just query `RoyaltiesInfo` and see if it returns a response for the same effect.

JakeHartnell: @orkunkl, any thoughts on this?

orkunkl (approved):

JakeHartnell (approved):

PR #28, opened 2021-11-09
v0.10.0

JakeHartnell:
Tagging `v0.10.0` version as per @ethanfrey's suggestion.

Also added a top level `cargo wasm` command.


the-frey (approved):
LGTM

ethanfrey:
published to crates.io

PR #29, opened 2021-11-10
Add some useful scripts

ethanfrey:
For updating the version on all packages at once:`./scripts/set_version.sh 0.10.1`
And for publishing all crates (for me until we add more owners to the crates.io repos)


the-frey (approved):
Can't test against crates obv but LGTM

JakeHartnell (approved):

PR #33, opened 2021-12-10
Move package/helper under cw721-base

orkunkl:
I was trying to use `Cw721Contract` inside another contracts tests.
But `cw721_base::ExecuteMsg` and `cw721::ExecuteMsg` are different types and includes `MintMsg`.
The easiest solution I found is moving the helper to cw721-base and using its messages.


orkunkl:
Changes merged in https://github.com/CosmWasm/cw-nfts/pull/34

PR #32, opened 2021-12-01
Implement approval query; fix helper queries

orkunkl:
This PR implements missing `approval()` query.

PR #34, opened 2021-12-13
Rename all_approvals to operators; Fix approval queries

orkunkl:
In my previous PR https://github.com/CosmWasm/cw-nfts/pull/32, I implemented approval query. Well I thought implemented it.
The problem is I read the logic above the code which was named `all_approvals` and then used the same query for individual approvals. But the logic was different. The `all_approvals` naming is not very explanatory for the logic. So I renamed it in this PR. fixed the approval query and added a query to gel all approvals of a token.


the-frey (approved):
Renaming makes sense. Don't know about operator, but unsure what a better word would be. Legally correct would be 'Proxy' but it is too overloaded in common tech usage :(

JakeHartnell (approved):
Makes a lot more sense to me. Thanks for this. :pray:

PR #40, opened 2021-12-29
Update cw 0.11

### LINKED ISSUE #39: Update to cw-plus v0.11
Mentioned here: https://github.com/CosmWasm/cw-nfts/pull/27#issuecomment-1002236943

It would be good to follow cw-plus release, and then push out cw721 v0.11.0 soon.
Happy if you move faster than cw-plus, but let's keep the latest version usable.

(This was release shortly before holidays, so no pressure to have it already, just think this is a good update to prioritise in the new year, and also meshes with importing cw1155)

orkunkl:
Resolves https://github.com/CosmWasm/cw-nfts/issues/39

### DIFF
--- a/contracts/cw721-base/src/query.rs
+++ b/contracts/cw721-base/src/query.rs
@@ -143,19 +143,15 @@ where
         let start = start_after.map(Bound::exclusive);
 
         let owner_addr = deps.api.addr_validate(&owner)?;
-        let res = self
+        let tokens: Vec<String> = self
             .tokens
             .idx
             .owner
             .prefix(owner_addr)
             .keys(deps.storage, start, None, Order::Ascending)
-            .take(limit);
-
-        let tokens = res
-            .into_iter()
-            .flatten()
-            .map(|addr| addr.to_string())
-            .collect();
+            .take(limit)
+            .map(|x| x.map(|addr| addr.to_string()))
+            .collect::<StdResult<Vec<_>>>()?;
 
         Ok(TokensResponse { tokens })
     }
--- a/contracts/cw721-base/src/query.rs
+++ b/contracts/cw721-base/src/query.rs
@@ -1,15 +1,15 @@
 use serde::de::DeserializeOwned;
 use serde::Serialize;
 
-use cosmwasm_std::{to_binary, Binary, BlockInfo, Deps, Env, Order, Record, StdError, StdResult};
+use cosmwasm_std::{to_binary, Addr, Binary, BlockInfo, Deps, Env, Order, StdError, StdResult};
 
-use cw0::maybe_addr;
 use cw721::{
     AllNftInfoResponse, ApprovalResponse, ApprovalsResponse, ContractInfoResponse, CustomMsg,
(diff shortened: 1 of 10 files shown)


contracts/cw721-base/src/query.rs
```diff
@@ -143,17 +143,20 @@ where
...
             .idx
             .owner
             .prefix(owner_addr)
             .keys(deps.storage, start, None, Order::Ascending)
-            .take(limit)
+            .take(limit);
+
+        let tokens = res
+            .into_iter()
+            .flatten()
```
ethanfrey: This is incorrect.
By flattening, you just filter out all Err() variants and unwrap the Ok() variant: https://doc.rust-lang.org/src/core/result.rs.html#1648-1664

We want an error if any fail, not silent ignoring

contracts/cw721-base/src/query.rs
```diff
@@ -143,17 +143,20 @@ where
...
         let owner_addr = deps.api.addr_validate(&owner)?;
-        let pks: Vec<_> = self
+        let res = self
             .tokens
             .idx
             .owner
             .prefix(owner_addr)
             .keys(deps.storage, start, None, Order::Ascending)
-            .take(limit)
+            .take(limit);
```
ethanfrey: Proper would be...

```rust
.take(limit)
.map(|x| x.map(|addr| addr.to_string())
.collect::StdResult<Vec<_>>()?;
```

ethanfrey: And with `let tokens: Vec<String> = self.tokens...` on the top

contracts/cw721-base/src/query.rs
```diff
@@ -271,10 +275,10 @@ where
     }
 }
 
-fn parse_approval(item: StdResult<Record<Expiration>>) -> StdResult<cw721::Approval> {
-    item.and_then(|(k, expires)| {
-        let spender = String::from_utf8(k)?;
-        Ok(cw721::Approval { spender, expires })
+fn parse_approval(item: StdResult<(Addr, Expiration)>) -> StdResult<cw721::Approval> {
+    item.map(|(spender, expires)| cw721::Approval {
```
ethanfrey, line 275: nice

contracts/cw721-base/src/state.rs
```diff
@@ -123,7 +123,7 @@ where
     T: Serialize + DeserializeOwned + Clone,
 {
     // pk goes to second tuple element
-    pub owner: MultiIndex<'a, (Addr, Vec<u8>), TokenInfo<T>>,
+    pub owner: MultiIndex<'a, Addr, TokenInfo<T>, Addr>,
```
ethanfrey, line 125: Maybe update the comment above?

ethanfrey (commented):
In general looks good.
`flatten().collect()` over a Result is not a good replacement for `collect::StdResult<Vec<_>>()`, it just silently ignores errors

ethanfrey (approved):
Looks good

PR #45, opened 2022-01-26
Allow inheritors of base contract to set their version correctly

### LINKED ISSUE #44: Invalid contract name for cw721-metadata-onchain
Currently metadata-onchain contract returns same cw721-base package name, but should be return its [crates name](https://crates.io/crates/cw721-metadata-onchain) : `crates.io:cw721-metadata-onchain`.

https://github.com/CosmWasm/cw-nfts/blob/d1a36319da1a432e0dd5ffe5c7dbb2e03a6777b6/contracts/cw721-base/src/execute.rs#L14

Maybe contrac_name should be set as contract_version and take name from `CARGO_PKG_NAME` 
https://github.com/CosmWasm/cw-nfts/blob/d1a36319da1a432e0dd5ffe5c7dbb2e03a6777b6/contracts/cw721-base/src/execute.rs#L15

the-frey:
Fixes #44 

This is a starting point to discuss how to fix the issue that contracts wholesale inheriting the base instantiate message have the incorrect contract version info.

This could be a problem when migrating, for example.

This PR proposes to add an optional field that takes a struct of values from a child contract. If set to `None` it will just use the base contract version info.


the-frey:
~Hm, I lack the perms to inspect/re-run the wasm build error.~

~Error is:~

```
Compiling cosmwasm-std v1.0.0-beta4
error: aborting due to 50 previous errors

For more information about this error, try `rustc --explain E0658`.
error: could not compile `rkyv`
```
EDIT: Error will be fixed by merge of #48

orkunkl:
I had came across this issue recently. The solution I found was overwriting the version.
```rust
  let res = stake_cw20::contract::instantiate(deps.branch(), _env, _info, msg)?;
    set_contract_version(deps.storage, CONTRACT_NAME, CONTRACT_VERSION)
        .map_err(ContractError::Std)?;
    Ok(res)
```
Adding a new field to instantiate makes things confusing.
Above looks like a simpler solution.

the-frey:
Hm, I don't disagree, exactly. I think it's a toss-up. My gut says I would rather have a clear parameter on the base contract, because then it's explicit and self-documenting.

BUT that means that it's effectively a breaking change to the API.

That said there will be breaking changes to the API in future, so perhaps that's not the end of the world.

orkunkl:
Agree with you on the need for passing versions down the contracts, but just passing a `VERSION` looks like just a patch. Maybe we can introduce a new variable to instantiate API called `ContractInfo` (or `Context`) where contract build context-related info is passed down? @ethanfrey

ethanfrey:
This does not belong at the cosmwasm-std level. And should be fixed in a library above it.

I don't know if this is the best idea, but let's try a few and see what the ecosystem convenes on.

I typically just call the super::instantiate and then call set_contract_version (again) in the "subclass".

It is 2 writes rather than 1, but only on instantiate and makes things simpler

the-frey:
So you write a custom instantiate call anyway.

Yes I guess that achieves the same thing without changing the core API so maybe this is more a documentation issue than anything else.

My main concern here is that people are deploying these contracts 'as is' so @orkunkl would you be happy for me to update this PR to use the double call approach that you and Ethan prefer?

orkunkl:
Best to update the PR.

JakeHartnell:
> I had came across this issue recently. The solution I found was overwriting the version.
> 
> ```rust
>   let res = stake_cw20::contract::instantiate(deps.branch(), _env, _info, msg)?;
>     set_contract_version(deps.storage, CONTRACT_NAME, CONTRACT_VERSION)
>         .map_err(ContractError::Std)?;
>     Ok(res)
> ```
> 
> Adding a new field to instantiate makes things confusing. Above looks like a simpler solution.

I think this approach is the way to go. Will make a new PR.

PR #47, opened 2022-01-28
Approval query returns positive for Owner

orkunkl:
In an NFT project i had this case: 
```rust
        // check if sender has approval
        nft_contract
            .approval(
                &deps.querier,
                token.id.clone(),
                info.sender.clone().into_string(),
                None,
            )
            .map_err(|_e| ContractError::Unauthorized {})?;

```
Where I only want to check if sender has permission to operate nft.
If the sender is `owner`, this query returns unauthorized. In order to make this work, I need to run one more owner query which makes things complicated and expensive.

### DIFF
--- a/contracts/cw721-base/src/contract_tests.rs
+++ b/contracts/cw721-base/src/contract_tests.rs
@@ -335,6 +335,26 @@ fn approving_revoking() {
         .execute(deps.as_mut(), mock_env(), minter, mint_msg)
         .unwrap();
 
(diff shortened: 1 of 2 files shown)


the-frey:
Looks good - suggest we merge https://github.com/CosmWasm/cw-nfts/pull/48 first, then the wasm build will succeed here and https://github.com/CosmWasm/cw-nfts/pull/45

JakeHartnell (approved):

PR #49, opened 2022-02-06
[cw721-fixed-price] Fix receiving cw20 msg

axvn:
Cw721-fixed-price contract raise deserialization exception on receiving cw20 tokens, because the message received is wrapped in `receive {..}`, not `cw20_receive_msg {..}`

> Cw20ReceiveMsg should be de/serialized under `Receive()` variant in a ExecuteMsg

[Ref](https://github.com/CosmWasm/cw-plus/blob/main/packages/cw20/src/receiver.rs#L6)


JakeHartnell (approved):
Great to have this fixed, thanks for the PR!

orkunkl (approved):
Awesome thanks!

PR #50, opened 2022-02-07
add pagination for cw721 queries to enable easier access to data

wrjanan:
Making use of cw-plus storage functionality skip for all queries which allow paging.

This is a breaking change in smart contracts as functions that query now require an extra `Option<u32>`
However for frontend it is non-breaking as request is an optional attribute.

This does not change any of the current implementation, merely exposes an alternative way of pagination as opposed to the `start_after` field. We can now page by page index.


JakeHartnell (dismissed):
LGTM, just a small lint issue in CI.

wrjanan:
There is a linked PR to cw-utils which includes paging struct, which will be used to reduce the number of variables in this function.

https://github.com/CosmWasm/cw-plus/pull/650

This PR will fix this linting issue, and open up standardization for paging to be used across repositories.

the-frey:
Yep, @wrjanan can you fix the linting issue? Then I am happy to give this the thumbs-up

wrjanan:
@the-frey 
Yes, I can fix the linting issue, the proper fix references the PR https://github.com/CosmWasm/cw-plus/pull/650?
This PR includes a fix for this linting issue.

PR adds a paging struct.

Which can be used for all requests that require paging, which is breaking, and it fixes this linting issue.

If we do not wish to create a breaking change, we can ignore the clippy lint particularly for this function. 

I am more towards adding a proper fix, hence created the PR in cw-plus for a paging struct. Since this paging struct should be something that can be easily referenced globally rather than specific to cw-nfts.

the-frey:
@orkunkl / @ethanfrey what do you think? Generally I'd go with upstream fix but there's the breaking/non breaking question as I understand it. 

Might be better to let contracts handle their own upgrade based on when they roll up their version of base & change any of their own custom logic?

Pros and cons to each approach I guess. Apols @wrjanan if I have misunderstood

wrjanan:
Add clippy ignore, this way the change will be a non-breaking change.

the-frey (approved):
Tentative thumbs-up from me, I think paging is pretty useful. This non-breaking version/keeping the same API shape is probably desirable as well IMO

shanev (commented):
An alternative pagination strategy is just extra bloat in a contract imho. Pagination is just a frontend convenience so the contract providing at least one way to do it is sufficient. Complex paginated queries are typically performed at the indexer / API level anyway. The only requirement of contract queries should be the provide the data in the simplest way possible without adding extra size to code.

This paginated style deviates from all the other contracts as well so it loses consistency.

orkunkl:
PR brings API breaking changes and requires maintenance on all contracts using this.
I am not sure if this new change is worth breaking the API. 
If you need the query, best to provide this with a new execution handler.

PR #64, opened 2022-05-26
Fix TokenIndexes MultiIndex

humanalgorithm:
The Multiindex for TokenIndexes works by accident and not by design

The key on a token index should be a String and not an Addr
The conversion of Addr to String is why existing setup does not fail any unit tests

An example usage below:  
If we have this data
```
"1": {owner: <addr1>, name: name1} 
"2": {owner: <addr1>, name: name2} 
"3": {owner: <addr2>, name: name3} 
```
The type structure being 
```
Map {
String: TokenInfo {
       "owner": <Addr>
       "data": <somedata>
    }, 
String: TokenInfo {
        "owner": <Addr>
        "data": <somedata>
    }, 
    ...
 }
 ```
To uniquely identify each entry, I need to differentiate by the key, which in this case is a String
The purpose of the index, is to get token ids by providing an owner address. So, 
input: `<addr1>` 
output: `["1", "2"]`

so the index is by owner <addr1> and then I get back the keys associated with it ["1", "2"]

Note that an owner (addr) can have multiple token ids (string)
So index by owner, is similar to tokens "group by" owner

### DIFF
--- a/contracts/cw721-base/src/query.rs
+++ b/contracts/cw721-base/src/query.rs
@@ -160,7 +160,6 @@ where
             .prefix(owner_addr)
             .keys(deps.storage, start, None, Order::Ascending)
             .take(limit)
-            .map(|x| x.map(|addr| addr.to_string()))
             .collect::<StdResult<Vec<_>>>()?;
 
         Ok(TokensResponse { tokens })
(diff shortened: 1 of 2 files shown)


shanev (approved):
Great catch! :P

orkunkl (approved):
Nice one!

orkunkl (approved):
Good catch! I made the change back when the development speed was fast. 
it is funny that,`Addr` is an abstraction to allow different address format validation, but in storage, it is simply an utf8 string. Lesson learned!

the-frey (approved):
Nice, good spot.

PR #67, opened 2022-07-18
Bump Max Query Limit

JakeHartnell:
It's a little too conservative. We should test this, but I think it will be fine...


0xekez:
Agree we should test this. Should make sure to do it on mainnet as well. In [my experience](https://github.com/DA0-DA0/dao-contracts/blob/c37d7d76713e27f2ba86b41b31810b8279022237/contracts/cw-proposal-single/src/contract.rs#L193-L206) testnet and mainnet gas limit differences can be tricky.

jhernandezb:
Worth to mention that gas for querying depends on per node configuration and not a consensus config, so nodes that don't want this level of exposure can still reduce the query gas limit. 

EDIT: This MAX_LIMIT is optional the client can choose a lower value.

PR #77, opened 2022-08-05
Fix trait to be compatible for custom modules

dadamu:
Currently, `CustomMsg` isn't the `cosmwasm-std` version and `CustomQuery` is not imported, it makes the current version is not compatible for custom modules, like `x/profiles` in Desmos chain. In addition, any messages to the contract should not be `CustomMsg` since they are not for the custom modules, they should be `DeserializeOwned`, instead. 
This PR intends to fix the issues mentioned above.


contracts/cw721-base/src/execute.rs
```diff
@@ -14,16 +16,17 @@ use crate::state::{Approval, Cw721Contract, TokenInfo};
 const CONTRACT_NAME: &str = "crates.io:cw721-base";
 const CONTRACT_VERSION: &str = env!("CARGO_PKG_VERSION");
 
-impl<'a, T, C, E, Q> Cw721Contract<'a, T, C, E, Q>
+impl<'a, T, E1, E2, C, Q> Cw721Contract<'a, T, E1, E2, C, Q>
 where
     T: Serialize + DeserializeOwned + Clone,
+    E1: DeserializeOwned,
+    E2: DeserializeOwned,
```
dadamu: `E1` is the execution message and `E2` is the execution query message to the contract. They are the messages to the contract but not to the chain custom modules, so they are set as `DeserializeOwned`.

contracts/cw721-base/src/execute.rs
```diff
@@ -14,16 +16,17 @@ use crate::state::{Approval, Cw721Contract, TokenInfo};
...
     C: CustomMsg,
-    E: CustomMsg,
-    Q: CustomMsg,
+    Q: CustomQuery,
 {
     pub fn instantiate(
         &self,
-        deps: DepsMut,
+        deps: DepsMut<Q>,
```
dadamu: Assigning `CustomQuery` to all the `Deps` in order to make communication to the custom modules available.

contracts/cw721-base/src/execute.rs
```diff
@@ -14,16 +16,17 @@ use crate::state::{Approval, Cw721Contract, TokenInfo};
...
-impl<'a, T, C, E, Q> Cw721Contract<'a, T, C, E, Q>
+impl<'a, T, E1, E2, C, Q> Cw721Contract<'a, T, E1, E2, C, Q>
 where
     T: Serialize + DeserializeOwned + Clone,
+    E1: DeserializeOwned,
+    E2: DeserializeOwned,
     C: CustomMsg,
-    E: CustomMsg,
-    Q: CustomMsg,
+    Q: CustomQuery,
```
0xekez: This is going to need to get rebased on this change which also changes these generic names: https://github.com/CosmWasm/cw-nfts/pull/75

How would you feel about giving `Q` here a more descriptive name? `DepsQueryExt`?

contracts/cw721-base/src/execute.rs
```diff
@@ -77,16 +80,16 @@ where
 }
 
 // TODO pull this into some sort of trait extension??
-impl<'a, T, C, E, Q> Cw721Contract<'a, T, C, E, Q>
+impl<'a, T, E1, E2, C, Q> Cw721Contract<'a, T, E1, E2, C, Q>
```
0xekez:
```suggestion
impl<'a, T, E1, E2, C, Q> Cw721Contract<'a, T, E1, E2, C, Q = Empty>
```

How about giving this the same default value as [CosmWasm does](https://docs.rs/cosmwasm-std/1.0.0/src/cosmwasm_std/deps.rs.html#26-30)?

contracts/cw721-base/src/helpers.rs
```diff
@@ -11,19 +12,32 @@ use serde::de::DeserializeOwned;
 use serde::{Deserialize, Serialize};
 
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
-pub struct Cw721Contract<Q: CustomMsg, E: CustomMsg>(
+pub struct Cw721Contract<
+    E1: Serialize + DeserializeOwned,
+    E2: Serialize + DeserializeOwned,
+    C: CustomMsg,
+    Q: CustomQuery,
```
0xekez:
```suggestion
    ModuleLevelCustomMsg: CustomMsg = Empty,
    ModuleLevelCustomQuery: CustomQuery = Empty,
```

0xekez (commented):
Great idea to include this! Just a comment about giving the new generics proper names and setting them to the same defaults `cosmwasm_std` does.

contracts/cw721-base/src/execute.rs
dadamu: I would name `Q` into `ModuleQuery` and `C` into `ModuleMsg`. What do you think?

contracts/cw721-base/src/execute.rs
dadamu: Yeah, it would be great adding default on its struct.

dadamu:
Rebased.

dadamu:
Updated to `main` branch, any updates for it? @0xekez

dadamu:
Updated to main branch, any updates for it? @0xekez

PR #27, opened 2021-11-09
Add CW-1155 to implement EIP 1155

sgoya:
This is a first pass at implementing cw-1155 for cw-nft repo.

PR #78, opened 2022-08-18
Add cw1155 and cw1155-metatdat-onchain

ALPAC-4:
[Cw1155 PR](https://github.com/CosmWasm/cw-nfts/pull/27) seems stop update. So I made new PR for adding cw1155

Most of the codes are based on [cw-plus repository](https://github.com/CosmWasm/cw-plus/tree/v0.13.4)

What is Added?

1. Can add `token_uri` and `extension` (metadata) when the first mint.
2. Add balance query (`all_balances`) by `token_id`
3. Add `NumTokens` query
4. Add `Minter` query

What is Removed?
1. Remove `batch_mint`


shanev (commented):
I think it makes sense for cw1155 to live in this repo. Happy to take another look after it has been updated to the latest versions and modernized (i.e: uses query response types, `cw_serde` macro, etc.).

giorgionocera:
Hi! I was rapidly looking at this contract to provide some support.
I'm sorry because I don't have much time right now to help with the code; I noticed something I think is wrong.
https://github.com/ALPAC-4/cw-nfts/blob/9e35f1ebacf0db34b2d9cad5c078c0d99cde6a24/contracts/cw1155-base/src/execute.rs#L119

Here, the token increment happens only if the token does not already exist.
So when you mint a token after the first mint, this quantity is not updated and the state is corrupted.
You can try to modify the tests to see the error, for example by minting the token at two different moments in time and then trying to access the balance.

What do you think?
I hope this could help.

packages/cw1155/src/event.rs
```diff
@@ -0,0 +1,41 @@
+use cosmwasm_std::{attr, Response, Uint128};
+use cw_utils::Event;
```
webmaster128, line 2: Could you try removing the dependency on this Event trait? I think we should remove it from cw_utils (https://github.com/CosmWasm/cw-utils/issues/17).

You probably can design this simpler without a trait but if you really need/want it, it's better to have it in here.

Thanks!

dirtyshab:
development of cw1155 is being continued here: https://github.com/CosmWasm/cw-nfts/pull/162

PR #79, opened 2022-09-13
Bump cosmwasm to v1.1, tidy up dependencies, update schema, update CI

larry0x:
## Changes

* Versioning
  - Bump Rust edition to 2021
  - Bump `cosmwasm-std` to v1.1
  - Bump `cw-plus` packages to `v0.14`
  - Bump all contracts and packages in this repo to `v0.14` to be consistent with cw-plus

* CI
  - Bump Rust version used in CI to v1.60.0 (was 1.58.1)
  - Bump cosmwasm-vm used in CI to v1.1.0 (was 1.0.0-beta8)

* Dependencies
  - Order all dependencies alphabetically
  - Removed a few unused dependencies
  - Moved a few dependencies that are only used in tests to dev-dependencies

* Schema
  - Replace the derive statement with the new `cw_serde` decorator, and regenerate schema accordingly:

```diff
- #[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
- #[serde(rename_all = "snake_case")]
+ #[cw_serde]
```

* Imports
  - in `cw721-base/src/helpers.rs` it imports the wrong `PhantomData`:

```diff
- use serde::__private::PhantomData;
+ use std::markers::PhantomData;
```

I assume this is because someone used editor auto-completion for imports...

## Notes

* The most powerful feature provided by cw v1.1, the query response derivation, does not work with cw721-base because it doesn't support query msgs that contain generics. I suggest removing `QueryMsg::Extension`.
* For types that contain generics, it is necessary to rename them when generating schemas. See [this example](https://github.com/CosmWasm/cw-nfts/blob/v0.13.4/contracts/cw721-base/examples/schema.rs#L21-L25). That is, you want `query_msg.json` instead of `query_msg_for__cw2891_query_msg.json`. The latter won't be picked up by the ts-codegen so is kinda useless.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -225,7 +225,7 @@ jobs:
       - run:
           name: Install check_contract
           # Uses --debug for compilation speed
-          command: cargo install --debug --version 1.0.0-beta8 --features iterator --example check_contract --locked -- cosmwasm-vm
+          command: cargo install --debug --version 1.1.0 --features iterator --example check_contract --locked -- cosmwasm-vm
       - save_cache:
           paths:
             - /usr/local/cargo/registry
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,7 +21,7 @@ workflows:
 jobs:
   contract_cw721_base:
     docker:
-      - image: rust:1.63.0
+      - image: rust:1.60.0
     working_directory: ~/project/contracts/cw721-base
     steps:
       - checkout:
@@ -31,7 +31,7 @@ jobs:
           command: rustc --version; cargo --version; rustup --version
       - restore_cache:
           keys:
-            - cargocache-cw721-base-rust:1.63.0-{{ checksum "~/project/Cargo.lock" }}
+            - cargocache-cw721-base-rust:1.60.0-{{ checksum "~/project/Cargo.lock" }}
       - run:
           name: Unit Tests
           environment:
@@ -53,11 +53,11 @@ jobs:
           paths:
             - /usr/local/cargo/registry
             - target
-          key: cargocache-cw721-base-rust:1.63.0-{{ checksum "~/project/Cargo.lock" }}
+          key: cargocache-cw721-base-rust:1.60.0-{{ checksum "~/project/Cargo.lock" }}
 
   contract_cw721_metadata_onchain:
     docker:
-      - image: rust:1.63.0
+      - image: rust:1.60.0
     working_directory: ~/project/contracts/cw721-metadata-onchain
     steps:
       - checkout:
@@ -67,7 +67,7 @@ jobs:
           command: rustc --version; cargo --version; rustup --version
       - restore_cache:
           keys:
-            - cargocache-cw721-metadata-onchain-rust:1.63.0-{{ checksum "~/project/Cargo.lock" }}
+            - cargocache-cw721-metadata-onchain-rust:1.60.0-{{ checksum "~/project/Cargo.lock" }}
       - run:
           name: Unit Tests
           environment:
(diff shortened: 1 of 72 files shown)


webmaster128:
> * For dependency version numbers, only keep the first two digits (e.g. `v1.x.y` --> `v1.x`). The 3rd digit marks non-breaking changes, so no need to hardcode it. Imo, better leaving it as a wildcard and let Rust compiler determine the most appropriate versions to use.

There can be good reason to do that. If you write "1.2.3" that means 1.2.0-1.2.2 are too low. Those versions might contain a bug you want to prevent in the codebase. Using 1.2 only includes those patch releases. Anyways, probably not important here.

.circleci/config.yml
```diff
@@ -21,7 +21,7 @@ workflows:
 jobs:
   contract_cw721_base:
     docker:
-      - image: rust:1.58.1
+      - image: rust:1.63.0
```
webmaster128: By changing this, you change the **minimal** Rust compiler version supported here. Is this really what you want? Using something like 1.60.0 ensures to not disrupt devs that don't work on latest stable.

larry0x:
> > * For dependency version numbers, only keep the first two digits (e.g. `v1.x.y` --> `v1.x`). The 3rd digit marks non-breaking changes, so no need to hardcode it. Imo, better leaving it as a wildcard and let Rust compiler determine the most appropriate versions to use.
> 
> There can be good reason to do that. If you write "1.2.3" that means 1.2.0-1.2.2 are too low. Those versions might contain a bug you want to prevent in the codebase. Using 1.2 only includes those patch releases. Anyways, probably not important here.

my understanding is that if i write 1.2, the compiler should automatically choose the latest, which is 1.2.3 in this example, unless there are conflicts making it unable to choose 1.2.3

contracts/cw2981-royalties/src/lib.rs
```diff
@@ -3,28 +3,28 @@ pub mod query;
...
 pub use cw721_base::{ContractError, InstantiateMsg, MintMsg, MinterResponse};
 
+use crate::msg::Cw2981QueryMsg;
+
 // Version info for migration
 const CONTRACT_NAME: &str = "crates.io:cw2981-royalties";
 const CONTRACT_VERSION: &str = env!("CARGO_PKG_VERSION");
 
-#[derive(Serialize, Deserialize, Clone, PartialEq, JsonSchema, Debug, Default)]
+#[cw_serde]
```
webmaster128, line 18: `Default` missing here as well, but maybe not needed

webmaster128 (dismissed):
I'm not familiar with this codebase but diff LGTM

webmaster128:
> my understanding is that if i write 1.2, the compiler should automatically choose the latest, which is 1.2.3 in this example, unless there are conflicts making it unable to choose 1.2.3

In the average case yes. But you still get lower version if:
- You have 1.2.2, 1.2.1, 1.2.0 in the lockfile already
- Some other crate pins version to e.g. "=1.2.2"

contracts/cw2981-royalties/src/lib.rs
larry0x: I don’t think this class should even have a Default trait. The default would be an attribute where both the key and the value are empty strings, which doesn’t make a lot of sense.

Another thing is that this class is repetitive code - it’s defined both in cw-metadata-onchain and in this contract. Imo it should be extracted to a standalone crate, which is exactly what we did at Stargaze: https://github.com/public-awesome/launchpad/tree/main/packages/sg-metadata

.circleci/config.yml
larry0x: Yeah good point, will change

larry0x:
> > my understanding is that if i write 1.2, the compiler should automatically choose the latest, which is 1.2.3 in this example, unless there are conflicts making it unable to choose 1.2.3
> 
> In the average case yes. But you still get lower version if:
> 
> * You have 1.2.2, 1.2.1, 1.2.0 in the lockfile already
> * Some other crate pins version to e.g. "=1.2.2"

Good point, will change back

contracts/cw2981-royalties/src/lib.rs
webmaster128: Alright, thanks for checking. True, many types should not have a Default (we are not in Go, right ;))

larry0x:
@webmaster128 I have downgraded the Rust version used in CI to 1.60.0 and use the full version numbers (1.x.y) for dependencies

webmaster128 (approved):
LGTM.

But I don't know who is codeowner here and who merged things.

JakeHartnell (approved):
LGTM! Thanks!

PR #80, opened 2022-09-14
Update cw-plus packages to 0.15.0

ueco-jb:

PR #81, opened 2022-09-15
Introduce usage of `schema-derive`

ueco-jb:
follow up of https://github.com/CosmWasm/cw-nfts/pull/80

This one is currently kind of blocked because of bug in new schema macro generator which takes generic arguments incorrectly.


ethanfrey:
> This one is currently kind of blocked because of bug in new schema macro generator which takes generic arguments incorrectly.

I merged the first PR (0.15 upgrade).
Can you make an issue on cosmwasm_std referencing this issue? (And link it here)

Unless there are objections, I would tag v0.15.0 now and we can do v0.15.1 with the better schema.

contracts/cw721-base/src/bin/schema.rs
```diff
@@ -0,0 +1,11 @@
+use cosmwasm_schema::write_api;
+
+use cw721_base::{ExecuteMsg, InstantiateMsg, QueryMsg};
+
+fn main() {
+    write_api! {
+        instantiate: InstantiateMsg,
+        execute: ExecuteMsg,
+        query: QueryMsg,
```
webmaster128: Can you do this instead (which is the same that was generated before)?:

```suggestion
        query: QueryMsg<Empty>,
```

PR #89, opened 2022-10-19
Use cargo workspace feature

larry0x:
Use Rust 1.64's new [cargo workspace](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table) feature to manage package info and dependencies.

NOTES:
- Currently CI fails because there is a clippy warning, which is out-of-scope for this PR. This is fixed by #92 
- Since the workspace feature is only available at Rust >=1.64.0, adopting this change means all contributors to this repo must update their Rust installation to minimally 1.64.0
- The bash script `scripts/set_version.sh` is kinda useless after this PR

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,7 +21,7 @@ workflows:
 jobs:
   contract_cw721_base:
     docker:
-      - image: rust:1.60.0
+      - image: rust:1.64.0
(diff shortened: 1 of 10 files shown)


shanev (approved):
Needed this badly! Thank you @larry0x!

JakeHartnell (approved):

PR #99, opened 2022-11-11
Cw4973

hoanm:
# Cw4973 - Account-bound Tokens (ABT)

Cw4973 is implemented based on the description of [EIP-4973](https://eips.ethereum.org/EIPS/eip-4973). Some requirements in the description will be changed for suitable with CosmWasm characteristics.

In the contract, I also implement the signature based on [ADR 036](https://github.com/cosmos/cosmos-sdk/blob/main/docs/architecture/adr-036-arbitrary-signature.md) and signature verification based on [@cosmjs/amino](https://www.npmjs.com/package/@cosmjs/amino).

## Specification
Cw4973 is extended from Cw721-base with some modifications in execution section:
* All the `metadata` structures are unchanged.
* All the `query functions` and `query messages` are unchanged.
* All the `execute functions` and `execute messages` are removed.
* We add 3 following `execute functions` and `execute messages` to the contract:
    - `Give`
    - `Take`
    - `Unequip`


contracts/cw4973/src/state.rs
```diff
@@ -0,0 +1,42 @@
...
+    pub value: MsgSignDataValue,
+}
+
+#[cw_serde]
+pub struct Fee {
+    pub amount: Vec<u8>,
+    pub gas: String,
+}
+
+// TODO: the order of these fields is VERY IMPORTANT, DO NOT CHANGE IT
```
JakeHartnell, line 25:
```suggestion
// The order of these fields is VERY IMPORTANT, DO NOT CHANGE IT
```

Not a TODO?

JakeHartnell (approved):
NICE!

contracts/cw4973/src/lib.rs
```diff
@@ -0,0 +1,319 @@
...
+#[cfg(not(feature = "library"))]
+pub mod entry {
+    use super::*;
+
+    #[entry_point]
+    pub fn instantiate(
+        mut deps: DepsMut,
+        env: Env,
+        info: MessageInfo,
+        msg: InstantiateMsg,
```
0xekez, line 43: what do you think about making the `InstantiateMsg` an alias for `Cw721BaseInstantiateMsg`([ex](https://github.com/DA0-DA0/dao-contracts/blob/8d7ad103c9c4bc3a2477a4d4f50532ef573f4d83/contracts/external/dao-migrator/src/msg.rs#L14-L16))?

```
type InstantiateMsg = Cw721BaseInstantiateMsg;
```

that would remove the need for the first line of the function and imo helps convey the intent a little better.

contracts/cw4973/src/lib.rs
```diff
@@ -0,0 +1,319 @@
...
+                Ok(hex::encode(hash))
+            }
+        }
+        Ok(false) => Err(ContractError::InvalidSignature),
+        Err(_) => Err(ContractError::CannotVerifySignature),
+    }
+}
+
+// the get_hash funtion will concat the address of the sender, the address of the 'to', the uri of the nft and the hash of the string
+fn _get_hash(active: &str, passive: &str, uri: &str, chain_id: &str) -> Vec<u8> {
```
0xekez, line 225: could you expand a little on what these function arguments are? are active, passive, and uri intended to be additional metadata? i'm curious why you include chain id?

contracts/cw4973/src/lib.rs
```diff
@@ -0,0 +1,319 @@
...
+        .add_attribute("token_id", msg.token_id))
+}
+// OR: create new contract to handle the signing document is better
+// create signable structure from message and chain ID
+// @param signer: the address of the signer
+// @param message: the message to sign
+// @param chain_id: the chain id of the chain
+// @return: the signable structure
+// TODO: modify this function to specify the others fields of the signing document
+fn _get_sign_doc(signer: &str, message: &str) -> String {
```
0xekez, line 279: this and your types in state.rs are 🙏.

1. is it possible to write an integration test for this? using cosmjs from the readme would be great.
2. if you'd like, i think this signing logic would be a very helpful package to have. would you be interested in breaking it out into one?

0xekez (commented):
i love the naming of the api and signature verification. just a couple questions from a first past.

PR #101, opened 2022-11-22
Update msg.rs

joemonem:


JakeHartnell (dismissed):

larry0x (dismissed):

JakeHartnell:
Lol, need to update schemas to fix CI.

larry0x (approved):

larry0x:
@JakeHartnell I updated schema. Can you re-review?

Also, it'd be better if this repo can be configured such that we don't need to re-review after every minor change.

JakeHartnell (approved):

JakeHartnell:
> Also, it'd be better if this repo can be configured such that we don't need to re-review after every minor change.

Update. :handshake:

PR #104, opened 2022-12-07
Delete `MintMsg`

larry0x:
Current:

```rust
enum ExecuteMsg<T> {
    Mint(MintMsg<T>),
}

struct MintMsg<T> {
    pub token_id: String,
    pub owner: String,
    pub token_uri: Option<String>,
    pub extension: T,
}
```

New:

```rust
enum ExecuteMsg {
    Mint {
        token_id: String,
        owner: String,
        token_uri: Option<String>,
        extension: T,
    },
}
```

This is backward compatible because `ExecuteMsg` serializes to the same JSON.

### DIFF
--- a/contracts/cw2981-royalties/schema/cw2981-royalties.json
+++ b/contracts/cw2981-royalties/schema/cw2981-royalties.json
@@ -212,7 +212,40 @@
         ],
         "properties": {
           "mint": {
-            "$ref": "#/definitions/MintMsg_for_Nullable_Metadata"
+            "type": "object",
+            "required": [
+              "owner",
+              "token_id"
+            ],
+            "properties": {
+              "extension": {
+                "description": "Any custom extension used by this contract",
+                "anyOf": [
+                  {
+                    "$ref": "#/definitions/Metadata"
+                  },
+                  {
+                    "type": "null"
+                  }
+                ]
+              },
+              "owner": {
+                "description": "The owner of the newly minter NFT",
+                "type": "string"
+              },
+              "token_id": {
+                "description": "Unique ID of the NFT",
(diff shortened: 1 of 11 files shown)


JakeHartnell (approved):

Art3miX (approved):

larry0x:
Would love to know your thoughts on this design decision @shanev @jhernandezb @0xekez

0xekez:
i'm pro. will need to update a couple contracts if we land this, but that's fine. forward!

larry0x:
@JakeHartnell do you know why CI doesn't run?

yubrew:
In the new version does `ExecuteMsg` still need `<T>`?

```
enum ExecuteMsg {
    Mint {
        token_id: String,
        owner: String,
        token_uri: Option<String>,
        extension: T,
    },
}
```

larry0x:
> In the new version does `ExecuteMsg` still need `<T>`?
> 
> ```
> enum ExecuteMsg {
>     Mint {
>         token_id: String,
>         owner: String,
>         token_uri: Option<String>,
>         extension: T,
>     },
> }
> ```

`T` is for the extension data, for example cw-metadata-onchain puts the metadata in the extension.

shanev (approved):
Nice simplification. `MintMsg` is a relic from the past before extensions. Although, it's arguable which is the cleaner path for extensions, adding extensions to `MintMsg`, or specialized `MintMsg`s. But it seems like we've settled in on the former.

JakeHartnell:
Not sure what's up with CI, but ran all checks locally, and we should be good. Thanks for this @larry0x!

PR #106, opened 2023-01-25
Bump dependency versions

larry0x:
- Bump core CosmWasm packages (cw-plus, cw-storage-plus, cw-utils) to v1.0
- ~Only specify the first one or two positions in semvar. E.g. `cosmwasm-std 1.2` instead of `cosmwasm-std 1.2.0`. @apollo-sturdy recently told me that these actually mean exactly the same - if you run `cargo update`, cargo updates the package to the latest under `1.*.*` regardless whether you specify the version as `1` or `1.2` or `1.2.0`~
- Bump cw-nfts to v0.17 since this is a breaking change
- The "raw" schema folders are deleted (not committed to the repo)


peara:
Hi, although `cosmwasm-std 1.2.0` and `cosmwasm-std 1.2` are the same to cargo. `cosmwasm-std 1.2.1` is not. 
As stated in [The Cargo Book](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio) :

> The string "0.1.12" is a version requirement. Although it looks like a specific version of the time crate, it actually specifies a range of versions and allows [SemVer](https://semver.org/) compatible updates. An update is allowed if the new version number does not modify the left-most non-zero digit in the major, minor, patch grouping. In this case, if we ran cargo update -p time, cargo should update us to version 0.1.13 if it is the latest 0.1.z release, but would not update us to 0.2.0.

I have encountered issues because of this when cosmwasm-std upgraded from 1.0.0 directly to 1.1.0 not long ago. I think to prevent this, packages should be versioned started from x.x.1 instead of x.x.0. But it seems many packages are doing that and I have not used rust long enough to understand. Anyway, another workaround is to specific the long dependency format such as: `cosmwasm-std = ">=1.2.0, <1.3.0"`.

Please consider adding this to prevent unexpected errors.

larry0x:
> Hi, although `cosmwasm-std 1.2.0` and `cosmwasm-std 1.2` are the same to cargo. `cosmwasm-std 1.2.1` is not. As stated in [The Cargo Book](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio) :
> 
> > The string "0.1.12" is a version requirement. Although it looks like a specific version of the time crate, it actually specifies a range of versions and allows [SemVer](https://semver.org/) compatible updates. An update is allowed if the new version number does not modify the left-most non-zero digit in the major, minor, patch grouping. In this case, if we ran cargo update -p time, cargo should update us to version 0.1.13 if it is the latest 0.1.z release, but would not update us to 0.2.0.
> 
> I have encountered issues because of this when cosmwasm-std upgraded from 1.0.0 directly to 1.1.0 not long ago. I think to prevent this, packages should be versioned started from x.x.1 instead of x.x.0. But it seems many packages are doing that and I have not used rust long enough to understand. Anyway, another workaround is to specific the long dependency format such as: `cosmwasm-std = ">=1.2.0, <1.3.0"`.
> 
> Please consider adding this to prevent unexpected errors.

I'm now using the full version numbers for dependencies in `Cargo.toml`.

Makefile.toml
```diff
@@ -36,7 +36,7 @@ fi
 docker run --rm -v "$(pwd)":/code \
   --mount type=volume,source="$(basename "$(pwd)")_cache",target=/code/target \
   --mount type=volume,source=registry_cache,target=/usr/local/cargo/registry \
-  ${image}:0.12.9
+  ${image}:0.12.11
```
larry0x, line 39: Update rust-optimizer version

Makefile.toml
```diff
@@ -56,6 +56,8 @@ for d in contracts/*; do
     cd ../..
   fi
 done
+
+rm -rf contracts/**/schema/raw
```
larry0x, line 60: The raw schema folders don't need to be committed

JakeHartnell (approved):

shanev (approved):

shanev:
Should we bump the main version to v1.0.1 for parity with cw-plus? I don't think we need to since this is a different repo now.

larry0x:
> Should we bump the main version to v1.0.1 for parity with cw-plus? I don't think we need to since this is a different repo now.

I expect we will still see major API changes in the repo. I think we should only tag v1 here once we are confident there won't be any major API change here.

PR #107, opened 2023-01-30
Add query for approval of 1 operator

### LINKED ISSUE #88: There is no easy way to query for existence of an operator
For operators added using `ApproveAll`, the only way to see them is using `AllOperators` query.
However, that query is paginated which makes it hard to use from another contract.
I suggest that we add a query type to directly check the existence of a `<owner, operator>` pair so operators can check immediately if  their approval statuses.

peara:
Fix #88.
I added a new query `Operator(owner, operator, include_expired)` which returns an `ApprovalResponse` if there exists such valid operator for the given owner. Otherwise, it will return an NotFound error.
This query should allow contracts such as `Marketplace` or `GameManager` check their required permission to manage users' NFTs easily.


peara:
@larry0x @shanev @JakeHartnell Could you give some thought?

JakeHartnell (approved):
Cool! I'm in support of adding this. We'll see what the others say.

larry0x:
I'm supportive of adding this query.

Regarding the response type, considering the other query methods have their own response types:

```rust
#[cw_serde]
pub struct ApprovalResponse {
    pub approval: Approval,
}

#[cw_serde]
pub struct ApprovalsResponse {
    pub approvals: Vec<Approval>,
}

#[cw_serde]
pub struct OperatorsResponse {
    pub operators: Vec<Approval>,
}
```

In order to be consistent, I think the `Operator` query should also have its own response type:

```rust
#[cw_serde]
pub struct OperatorResponse {
    pub approval: Approval,
}
```

larry0x:
Off topic - I think the `AllOperators` query should be renamed to `Operators` to be consistent with the naming convention of other queries.

peara:
> Off topic - I think the `AllOperators` query should be renamed to `Operators` to be consistent with the naming convention of other queries.

Although, I agree with this, this is a breaking change. There are also `AllNftInfo` and `AllTokens` which should be considered as well.

shanev (approved):
Was thinking why this didn't exist the other day.

PR #109, opened 2023-02-04
Make cw721-base minter updatable via two-step ownership transfer.

0xekez:
This makes a breaking change to cw721-base in that it makes the `minter` field on `cw721_base::MinterResponse` optional.

Before:

```rust
pub struct MinterResponse {
    pub minter: String,
}
```

After:

```rust
pub struct MinterResponse {
    pub minter: Option<String>,
}
```

It also makes a breaking change to the `ContractError` enum in that it removes the `Unauthorized` variant in favor of cw-ownable's more descriptive `OwnershipError`.

It then extends the `cw721_base`, but not the `cw721` standard's, `ExecuteMsg` with
[`cw-ownable`](https://crates.io/crates/cw-ownable)'s `UpdateOwnership` enum variant, and extends the `QueryMsg` with `cw-ownable`'s `Ownership {}` query. Using these messages, the current minter on the contract may nominate a new minter, who may accept or reject this nomination.


0xekez:
@shanev , @JakeHartnell , @larry0x - i'd be interested in your feedback about this.

we'd like to use something like this in DAO DAO to allow NFT DAOs to [bootstrap themselves](https://github.com/DA0-DA0/dao-contracts/issues/642). ICS-721 would possibly also benefit from a more flexible ownership scheme as it could allow the owner to control rate-limits and other parameters relevant to their collection moving to other chains (where they could lose royalty protections, for example, when leaving Stargaze).

the change i've made here tries to limit API breakage by not renaming the `minter` field on the `InstantiateMsg`. we could further break the API by changing that field to `owner`. at the moment, $minter \iff owner$.

Cargo.toml
```diff
@@ -21,6 +21,7 @@ cw-utils        = "0.16.0"
 schemars        = "0.8.10"
 serde           = { version = "1.0.140", default-features = false, features = ["derive"] }
 thiserror       = "1.0.31"
+cw-ownable      = "0.4.0"
```
larry0x: Can you sort deps alphabetically

contracts/cw721-base/Cargo.toml
```diff
@@ -31,3 +31,4 @@ cw-storage-plus = { workspace = true }
 schemars        = { workspace = true }
 serde           = { workspace = true }
 thiserror       = { workspace = true }
+cw-ownable      = { workspace = true }
```
larry0x: Alphabetically

contracts/cw721-base/src/error.rs
```diff
@@ -1,13 +1,14 @@
 use cosmwasm_std::StdError;
+use cw_ownable::OwnershipError;
 use thiserror::Error;
 
 #[derive(Error, Debug, PartialEq)]
 pub enum ContractError {
-    #[error("{0}")]
+    #[error(transparent)]
```
larry0x, line 7: Is there any difference between `"{0}"` and `transparent`?

contracts/cw721-base/src/query.rs
```diff
@@ -207,6 +207,10 @@ where
             },
         })
     }
+
+    fn ownership(deps: Deps) -> StdResult<cw_ownable::Ownership<Addr>> {
```
larry0x: I think we should return `Ownership<String>` here. The idea is that contract APIs (i.e. messages and responses) should only contain unchecked types.

contracts/cw721-base/src/query.rs
```diff
@@ -217,10 +221,11 @@ where
     Q: CustomMsg,
 {
     pub fn minter(&self, deps: Deps) -> StdResult<MinterResponse> {
```
larry0x: It'd be API-breaking but I think we can just delete this query method

JakeHartnell:
> we could further break the API by changing that field to owner.

Damn, I really want to do this but feel it would be a pain for a lot of folks...

larry0x (approved):
Yay cw-ownable adoption!

larry0x:
> > we could further break the API by changing that field to owner.
> 
> Damn, I really want to do this but feel it would be a pain for a lot of folks...

Opinions from @shanev @jhernandezb should be the most important regarding API breakages... What do you guys think about:

- renaming the `minter` field in instantiate msg
- delete `QueryMsg::Minter` and use `QueryMsg::Ownership` instead

contracts/cw721-base/src/error.rs
0xekez: this is the way that the `thiserror` docs suggest passing through errors, though the actual serialization doesn't change. i feel this is a little more explicit and that's what i do in other contracts, though i don't have an opinion on what we use in this repo if we'd like to stick to `{0}`.

contracts/cw721-base/src/query.rs
larry0x: This probably doesn't matter anyways, since the result is serialized into JSON

larry0x:
@0xekez I think you can consider adding `rust-version = "1.65"` in cw721-base's `Cargo.toml` to enforce a minimum Rust version

(For context: cw-ownable uses a syntax only available since Rust 1.65)

shanev (approved):
Fully support this since Stargaze NFT DAOs will also need this to bootstrap themselves. My only concern is that use of the term `ownership` can be confused with NFT ownership.

jhernandezb:
I think it would be nice to have an example of how to migrate a contract from previous versions in case someone misses it and leaves a contract uninitialized.

larry0x:
> I think it would be nice to have an example of how to migrate a contract from previous versions in case someone misses it and leaves a contract uninitialized.

The way I suggest to do this is to create an `src/migrations` folder with this structure:

```plain
contracts/
└── cw721-base/
    └── src/
        └── migrations/
            ├── mod.rs
            ├── v0_16.rs
            ├── v0_17.rs
            ├── v1_0.rs
            └── v1_1.rs
```

Each `v*_*.rs` exports a single `migrate` function that contains the logic for migrating from previous version to the current version. For example, `v0_17.rs` will contain the logic for migrating v0.16 to v0.17.

To migrate a custom cw721 contract from v0.16 to v0.17, simply do:

```rust
#[entry_point]
fn migrate(deps: Deps, env: Env, msg: MigrateMsg) -> Result<Response, ContractError> {
    cw721_base::migrations::v0_17::migrate(deps, env, msg)?;
    Ok(Response::new())
}
```

larry0x:
I can imagine this would be tricky for Stargaze tho... Since each NFT collection is a separate contract, we have to migrate each one of them. It may be useful to create an sdk module that registers each collection, and provide a governance-gated message to migrate them all together.

jhernandezb:
> I can imagine this would be tricky for Stargaze tho... Since each NFT collection is a separate contract, we have to migrate each one of them. It may be useful to create an sdk module that registers each collection, and provide a governance-gated message to migrate them all together.

I governance proposal would be a nice one to have, currently creators have admin and can do the migration themselves through our studio UI. Your proposed solution sounds good to me

0xekez:
> @jhernandezb: I think it would be nice to have an example of how to migrate a contract from previous versions in case someone misses it and leaves a contract uninitialized.

good idea, i've added migration logic and added this to the README:

## Updating the minter

For contract versions $> 0.16$ the minter has been replaced by an
owner which is updatable via the two-step ownership transfer process
of [cw_ownable](https://crates.io/crates/cw-ownable). To retrieve the
owner, `QueryMsg::Ownership {}` may be executed.

`QueryMsg::Minter {}` has not been removed, though after version 0.16
the response type has made the minter field optional as it may be
unset. For all intents and purposes, whenever the word minter is used,
it means owner, and owner in turn means minter, $minter \iff owner$.

Before 0.16:

```rust
pub struct MinterResponse {
    pub minter: String,
}
```

After 0.16:

```rust
pub struct MinterResponse {
    pub minter: Option<String>,
}
```

NFTs on version 0.16 may upgrade via `MigrateMsg::From016 {}`. For an
example of doing so, see
[this](https://github.com/CosmWasm/cw-nfts/blob/zeke/updatable-minter/contracts/cw721-base/src/multi_tests.rs#L83)
integration test.

0xekez:
i'll merge this tomorrow if there are no more comments before then

Cargo.toml
```diff
@@ -8,16 +8,20 @@ license       = "Apache-2.0"
 repository    = "https://github.com/CosmWasm/cw-nfts"
 homepage      = "https://cosmwasm.com"
 documentation = "https://docs.cosmwasm.com"
+rust-version = "1.65"
```
larry0x: Align with other lines

```suggestion
rust-version  = "1.65"
```

contracts/cw721-base/README.md
```diff
@@ -30,6 +30,39 @@ or just create a custom contract as the owner and use that contract to Mint.
 
 If provided, it is expected that the _token_uri_ points to a JSON file following the [ERC721 Metadata JSON Schema](https://eips.ethereum.org/EIPS/eip-721).
 
+## Updating the minter
```
larry0x: I think this is better put in a separate `MIGRATING.md` which contains all the migration guides ordered by versions

contracts/cw721-base/README.md
```diff
@@ -30,6 +30,39 @@ or just create a custom contract as the owner and use that contract to Mint.
...
+
+For contract versions &\gt 0.16$ the minter has been replaced by an
```
larry0x: Suggest not using LaTeX, since not all markdown renders support it

contracts/cw721-base/README.md
```diff
@@ -30,6 +30,39 @@ or just create a custom contract as the owner and use that contract to Mint.
...
+```
+
+NFTs on version 0.16 may upgrade via `MigrateMsg::From016 {}`. For an
+example of doing so, see
+[this](https://github.com/CosmWasm/cw-nfts/blob/zeke/updatable-minter/contracts/cw721-base/src/multi_tests.rs#L83)
+integration test.
+
 ## Running this contract
 
 You will need Rust 1.44.1+ with `wasm32-unknown-unknown` target installed.
```
larry0x: This line probably should be updated, since cw-ownable only supports Rust 1.65+

contracts/cw721-base/src/execute.rs
```diff
@@ -125,6 +123,39 @@ where
...
+        action: cw_ownable::Action,
+    ) -> Result<Response<C>, ContractError> {
+        let ownership = cw_ownable::update_ownership(deps, &env.block, &info.sender, action)?;
+        Ok(Response::new().add_attributes(ownership.into_attributes()))
+    }
+
+    pub fn migrate(deps: DepsMut, msg: MigrateMsg) -> Result<Response<C>, ContractError> {
+        let ContractVersion { version, .. } = get_contract_version(deps.storage)?;
+        set_contract_version(deps.storage, CONTRACT_NAME, CONTRACT_VERSION)?;
+        match msg {
```
larry0x: Against this approach, because this way the contract will contain all the historical migration logics: e.g. a v0.18 contract will include both v0.17 -> v0.18 migration logics, but also that for v0.16 -> v0.17, etc. This way, the wasm binary size will keep growing as we add more and more migration logics to it.

Imo, migration logics should be included in a way that older, unused logics can be optimized away. Can you take a look at what I do in badges project and let me know what you think: https://github.com/steak-enjoyers/badges/tree/master/contracts/hub/src/upgrades

contracts/cw721-base/src/execute.rs
```diff
@@ -125,6 +123,39 @@ where
...
+                    Err(ContractError::WrongMigrateVersion(version))
+                } else {
+                    use cw721_base_016 as v16;
+                    let tract16 = v16::Cw721Contract::<T, C, E, Q>::default();
+                    let minter = tract16.minter.load(deps.storage)?;
+                    tract16.minter.remove(deps.storage);
+                    cw_ownable::initialize_owner(deps.storage, deps.api, Some(minter.as_str()))?;
+                    let ownership = cw_ownable::get_ownership(deps.storage)?;
+                    Ok(Response::new()
+                        .add_attribute("method", "migrate_from_016")
```
larry0x: Other execute method using `"action"` has the key here. I think we should be consistent here. Also suggest just naming this action as "migrate" and use attributes to indicate the specific migrations being migrated from/to:

```rust
Response::new()
    .add_attribute("action", "migrate")
    .add_attribute("from_version", "0.16")
    .add_attribute("to_version", "0.17")
```

contracts/cw721-base/README.md
larry0x: Like this: https://github.com/CosmWasm/cosmwasm/blob/main/MIGRATING.md

0xekez: great idea, thanks

contracts/cw721-base/src/execute.rs
0xekez: thanks for the thoughts, i read your earlier comment but already had this done, minus test updates. here's my argument for this approach:

what is our biggest concern for contract migrations? imo:

1. an accidental state key change in a newer version that causes state to go missing after migration. [ex](https://github.com/DA0-DA0/dao-contracts/pull/548#discussion_r1029936817).
2. a small change in the way types are serialized between versions (perhaps in a dependency) causes deserialization to fail post-migration. [ex](https://github.com/DA0-DA0/dao-contracts/pull/465#discussion_r958808276).

as the examples show, careful people mess these things up frequently.

in my view, this system is optimal then as it protects us against these in many ways.

1. we are guaranteed to load from the same hardcoded state key.
2. we are guaranteed to get the deserialization method used by the old version.

then, using the old import we can run unit tests against _exactly_ the old contract and perform state checks before and after (see the `mint_transfer_burn` call before and after migration in my integration test). we track contract size and upload cost over time in dao-contracts and i have not seen notable increases from adding these dependencies, i expect this is the rust linker removing the unused parts.

i do understand that this reduces the ease of multi-hop migrations. in dao-contracts we only maintain a `FromVersion {}` migration variant for one version. So, do do a multi-hop upgrade you'd need to migrate twice.

larry0x: Understand the concern, but my opinion is this: _each version of the contract should only contain one migration logic, from the immediate previous version to the current version._ E.g. contract 0.17 should only contain 0.16->0.17, contract 0.18 should only contain 0.17->0.18, so on.

Of course, when migrating 0.16->0.17, the contract must check cw2 that it's indeed being migrated from 0.16.

If you have a 0.16 contract and want to migrate to 0.18, you have to do this sequentially: first 0.16->0.17 using the 0.18 code_id, then 0.17->0.18 with the 0.18 code_id.

With this design we get both a small wasm binary size, and avoid the issue you're concerned about.

0xekez: @larry0x would you mind taking a look at this commit and seeing if it fits what you're saying? https://github.com/CosmWasm/cw-nfts/pull/109/commits/6b934b624353826afb706b0604c79b485936d53a

JakeHartnell (approved):

contracts/cw721-base/src/upgrades/v0_16.rs
```diff
@@ -0,0 +1,22 @@
...
+where
+    T: Serialize + DeserializeOwned + Clone,
+    Q: CustomMsg,
+    E: CustomMsg,
+{
+    let tract16 = v16::Cw721Contract::<T, C, E, Q>::default();
+    let minter = tract16.minter.load(deps.storage)?;
+    tract16.minter.remove(deps.storage);
+    cw_ownable::initialize_owner(deps.storage, deps.api, Some(minter.as_str()))?;
+    let ownership = cw_ownable::get_ownership(deps.storage)?;
```
larry0x: We can make `initialize_owner` return the ownership info, so that you don't need to `get_ownership` here. Thoughts? I can make a new release of cw-ownable if you want to do it in this PR.

```rust
let ownership = cw_ownable::initialize_owner(deps.storage, deps.api, Some(minter.as_str()))?;
```

contracts/cw721-base/src/upgrades/v0_16.rs
```diff
@@ -0,0 +1,22 @@
...
+    Ok(Response::new()
+        .add_attribute("action", "migrate_016_017")
```
larry0x:
```suggestion
        .add_attribute("action", "migrate")
        .add_attribute("from_version", "0.16.0")
        .add_attribute("to_version", "0.17.0")
```

larry0x (commented):
@0xekez The additions of `MIGRATING.md` and `v0_16.rs` look good! I left a two comments.

larry0x:
@0xekez @JakeHartnell I'd say this PR is ready to be merged! However, please don't cut a release for 0.17 yet. I say we include #106, perhaps also #107, and I also have a suggestion to how we do cw2 versioning in cw721-base which I'll do a PR shortly.

JakeHartnell:
> I say we include #106, perhaps also #107, and I also have a suggestion to how we do cw2 versioning in cw721-base which I'll do a PR shortly.

Completely agree.

contracts/cw721-base/src/upgrades/v0_16.rs
JakeHartnell: That would be sick!

0xekez: @larry0x https://github.com/steak-enjoyers/cw-plus-plus/pull/12 :)

larry0x: @0xekez Just released cw-ownable v0.5.0 with this change

PR #111, opened 2023-02-16
Validate royalty percentage when minting cw2981

peara:
## Problem 
At the moment, there is no validation of `royalty_percentage` when minting new token in `cw2981-royalties`.
As percentage over 100% is mostly nonsensical in the context of royalty, it should be checked to prevent potential errors.
For example, a marketplace contract can mistakenly send their own tokens instead of the buyer's because royalty amount will be larger than the price.

=> validate `royalty_percentage` to be smaller than equal to 100.


peara:
@JakeHartnell @shanev @larry0x Please check this PR.

shanev (approved):

contracts/cw2981-royalties/src/lib.rs
```diff
@@ -80,7 +80,40 @@ pub mod entry {
...
+            } => {
+                // validate royalty_percentage to be between 0 and 100
+                if let Some(Metadata {
+                    royalty_percentage: Some(royalty_percentage),
+                    ..
+                }) = extension
+                {
+                    // no need to check < 0 because royalty_percentage is u64
+                    if royalty_percentage > 100 {
+                        return Err(ContractError::Std(StdError::GenericErr {
```
larry0x: We should define our own error type instead of inheriting from cw721-base. Let me see if I can push a commit

larry0x: @peara I pushed a commit, please check: https://github.com/CosmWasm/cw-nfts/pull/111/commits/db45da19e387d779992f076101e6137d85c1d8bd

peara: Thank you, it looks much cleaner.

peara:
@larry0x Can we go?

larry0x (approved):

PR #112, opened 2023-02-28
v0.17 upgrade logic

larry0x:
- Rename `upgrades/v0_16.rs` to `upgrades/v0_17.rs` (I think the filename should be the version being migrated **to**, not the version being migrated **from**)
- Version check to make sure that correct contract is being migrated and it's being migrated from the correct version (using `cw2::assert_contract_version`; have to use our fork of cw2 before [our PR](https://github.com/CosmWasm/cw-plus/pull/858) is merged)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,9 @@ rust-version  = "1.65"
 [workspace.dependencies]
 cosmwasm-schema = "1.2.1"
 cosmwasm-std    = "1.2.1"
-cw2             = "1.0.1"
+# FIXME: change back to the "official" crates.io release after this PR is merged:
+# https://github.com/CosmWasm/cw-plus/pull/858
+cw2             = { git = "https://github.com/mars-protocol/cw-plus", rev = "1a3a944" }
 cw20            = "1.0.1"
 cw721           = { version = "0.17.0", path = "./packages/cw721" }
 cw721-base      = { version = "0.17.0", path = "./contracts/cw721-base" }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -294,7 +294,7 @@ checksum = "c80e93d1deccb8588db03945016a292c3c631e6325d349ebb35d2db6f4f946f7"
 dependencies = [
  "cosmwasm-schema",
  "cosmwasm-std",
- "cw2 1.0.1",
+ "cw2 1.0.1 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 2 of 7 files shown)


JakeHartnell (approved):
LGTM

0xekez (approved):
Nice improvement. Thanks!

0xekez:
@larry0x sorry for being so slow to review. should be able to rebase on main and merge this.

Cargo.toml
```diff
@@ -13,7 +13,9 @@ rust-version  = "1.65"
 [workspace.dependencies]
 cosmwasm-schema = "1.2.1"
 cosmwasm-std    = "1.2.1"
-cw2             = "1.0.1"
+# FIXME: change back to the "official" crates.io release after this PR is merged:
+# https://github.com/CosmWasm/cw-plus/pull/858
+cw2             = { git = "https://github.com/mars-protocol/cw-plus", rev = "1a3a944" }
```
Art3miX, line 18: Seems to be an overkill just to get the `assert_contract_version` helper: https://github.com/CosmWasm/cw-plus/pull/858/files#diff-a35715fa1fde56c42db0d8f2845ebc6a571a801699c0705f0080237820d96ee6R59-R83

Would suggest just to just put it in this contract until we have an "official" version of it.

Art3miX (commented):
Seems good.

Just little comment about the cw2 library, I think version assertion is something custom to each contract (why cw2 didn't create a function for it?), would defo suggest just copy paste it as a helper.

Cargo.toml
larry0x: @Art3miX My PR to cw2 referenced in the comment has been merged. Now we can use the main branch of cw2 instead of mars-protocol fork. Imo this is more elegant than including extra code in this repo which we need to remove later.

Art3miX: Oh ok, so this can be changed to the official cw2 now? 

I just didn't like the using some fork for a helper function just to revert back later to the previous package.

larry0x: We can change it to the main branch (use the latest commit) of the official cw-plus repo

shanev (approved):

Cargo.toml
humanalgorithm: Not sure what's going on here but when I go to the main branch of cw2 I am not seeing the assert_contract_version that is referenced here.

PR #120, opened 2023-04-22
cw721-template

### LINKED ISSUE #118: Make a template for starting a new basic cw721 project
Should work similar to: https://github.com/CosmWasm/cw-template

Contract would be a simple one with no extensions.

JakeHartnell:
Closes #118.

Need to write more contract templates... `cargo-generate` is pretty cool.


Art3miX:
I really like the templates idea, while this might seem like not needed, or can be easily achieved by simply copying a folder without generating a template, those sort of small developing improvements are exactly what cosmos is missing.

The only thing I don't like is the fact all of the code is in the `lib.rs` file, templates should be structurally as similar as possible to a working contract, in my opinion it will be much better to separate the lib.rs file into the corresponding files: errors.rs, msg.rs, contract.rs, tests.rs

JakeHartnell:
> I really like the templates idea, while this might seem like not needed, or can be easily achieved by simply copying a folder without generating a template, those sort of small developing improvements are exactly what cosmos is missing.
> 
> The only thing I don't like is the fact all of the code is in the `lib.rs` file, templates should be structurally as similar as possible to a working contract, in my opinion it will be much better to separate the lib.rs file into the corresponding files: errors.rs, msg.rs, contract.rs, tests.rs

I've been meaning to break it up but haven't found the time yet, that's a really good suggestion. Would be great to upstream a template to `sg721` as well.

Art3miX:
https://github.com/JakeHartnell/cw-nfts/pull/1

elix1er:
break it up!

PR #127, opened 2023-06-21
`cw721-sylvia-base`

### LINKED ISSUE #31: Extension is difficult & other long term design decisions
I think this is the starting point for a wider conversation. Or just tell me I'm imagining things. Either or.

FWIW I can use and extend this stuff fine now, hence why I will totally accept "nah, it's fine" as an answer.

However, I've found it a bit of an uphill struggle to get to grips with, and am still a little unclear on best practice for extending. I've seen several others taking similar or different approaches to this (on the query and the execute side). Generally, doing roughly what is in:

https://github.com/envoylabs/whoami/blob/main/src/lib.rs#L36 for execute and
https://github.com/CosmWasm/cw-nfts/blob/d79bd6f7803dd366430ade1b77edbeaea74946ca/contracts/cw2981-royalties/src/lib.rs#L79 for query

seems to be kinda how most folk do it (roughly).

I'm not experienced enough at Rust to make a perfect call on this, but my gut feeling is that the way CW721 base is implemented is:

- unapproachable for newer rust developers
- difficult to extend
- results in copy-paste code in new subprojects
- not actually saving much duplication

If I didn't have the Rust compiler to help, I'd be relatively lost.

I found myself just now needing to essentially copy-and-paste a big chunk of functionality, and I was thinking "how much of the base am I using and not overriding?" The way it works right now is as a container, and that's fine - because there is base behaviour in there that I _am_ using, but I'm not sure what it is buying me over a module containing functions that I can import as-and-when-needed, if I'm having to implement new functions in a module and dispatch to them, instead of bolting stuff to the side of the container. The default implementations of `query`, `execute` etc I suppose?

What I fear is that base behaviour will change but not many contracts will pick it up, because it's being overriden, at which point you have to ask whether this should be a lego-bricks library of helper functions with some enums (closer to how some of the other cw-plus contracts are implemented, I suppose).

I'm sticking my neck out a little bit here, so feel free to say that I'm just not good with traits and I'm missing something, but... yeah. I feel like this is probably difficult to decide via e.g. discord and we need to have a design discussion around this. Even if it's just "the-frey, you're wrong, but X _is_ an issue, let's talk about **that**."

Usage of these contracts is already exploding and soon we won't have the luxury to radically change `base`. 

Which also brings into focus discussions like #25 - how we extend, what we consider e.g. `base` and `plus` and how we want to make a modular code base that is ergonomic to use while still being safe.

I don't think any of these are easy questions, and I certainly don't want to point fingers - everybody has worked hard to get this library to the place it is today (thanks!) but I see the window for foundational design closing, to some degree.

Maybe this also ties into the wider standards conversation that's happening. We're building the plane on the way down, and that's fine, but a few hours and some discussion now could save us some heartache in the future if we're going to be designing and building modular CW contracts for the next few years around NFTs.

Sorry about the essay and apologies if it's unclear what I'm trying to say. It's been a long day of code and possibly I need to revisit and refine this later.

JakeHartnell:
Adds a similar contract to `cw721-base` but written with the [Sylvia](https://github.com/cosmwasm/sylvia) smart contract framework.

There are a few advantages to Sylvia over the extension based pattern we currently have:
- Custom Queries and Execute messages can be added at the top level. No more need to wrap them in `ExecuteMsg::Extension`.
- Easier to add new custom messages and queries
- Less interaction with traits, way less complicated types
- Easier to write tests! Sylvia + multi-test is super sweet

Have some open questions which I would love folks thoughts on if they have them:
- Do we still want to support extensions at all for token metadata? Or do we just kill them? (personally leaning towards the later)
- How should we organize this codebase? Took a stab based on some Sylvia examples, but open to suggestions.
- As we move forward, should we focus more on a Sylvia based approach? Or stick with the extension pattern we have (which is already well used).

Closes #31 hopefully someday :crossed_fingers:


the-frey:
Top line thought: this is useful to have in the repo. If Sylvia becomes more widely adopted, we should begin to provide a 'paved path'. If it doesn't, then it's either still useful documentation for somebody, or could be hypothetically removed later (although, personally once something is in, I think we should maintain it, see below).

> Do we still want to support extensions at all for token metadata? Or do we just kill them? (personally leaning towards the later)

Depending on the effort to do so, yes, because they are widely in production and it's probably better to offer alternatives rather than hard deprecate a given pattern.

> How should we organize this codebase? Took a stab based on some Sylvia examples, but open to suggestions.

No strong thoughts on this one as don't personally use Sylvia.

> As we move forward, should we focus more on a Sylvia based approach? Or stick with the extension pattern we have (which is already well used).

I guess as referred to in my first response I think folks that want to focus on Sylvia can put those contracts here so that other people can use them, but if people don't want to use Sylvia, then that's valid. Any framework usually has some drawbacks WRT flexibility and from my understanding the same is true here. 

There's another question of whether it's more approachable to beginners. My guess would be no, as there's a bit more upfront type stuff to parse in exchange for simpler definitions of data, but could be wrong about that!

JakeHartnell:
Added more tests, coverage should be decent by now. Will add a template for a basic sylvia cw721 contract after this is reviewed.

PR #144, opened 2023-11-21
Add cw721-receiver contract

chipshort:
This implements an example contract that can receive a cw721 token and then just succeed or error depending on the passed message.

Also fine if you don't want to merge. We just needed this for a wasmd test. In that case, just close it.


JakeHartnell:
This is really helpful! Have had to make similar contracts to this myself... maybe we need a `test-contracts` folder? Open to ideas from others but would be down to have some test contracts like this.

shanev (approved):

Art3miX:
> This is really helpful! Have had to make similar contracts to this myself... maybe we need a `test-contracts` folder? Open to ideas from others but would be down to have some test contracts like this.

I think having a `test-contracts` folder is smart, you almost always need a test contract if you work with cw721

contracts/cw721-receiver/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+# for more explicit tests, cargo test --features=backtraces
+backtraces = ["cosmwasm-std/backtraces"]
+# use library feature to disable all instantiate/execute/query exports
+library = []
+
+[dependencies]
+cosmwasm-schema = { workspace = true }
+cosmwasm-std = { workspace = true }
+cw-storage-plus = { workspace = true }
+cw2 = { workspace = true }
```
Art3miX: Not needed as you don't use them in the contract?

contracts/cw721-receiver/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+cw721 = { workspace = true }
+cw721-base = { workspace = true, features = ["library"] }
```
Art3miX: Only used in tests, so can move to the dev-deps?

contracts/cw721-receiver/tests/multitest.rs
```diff
@@ -0,0 +1,76 @@
...
+            &base_msg::ExecuteMsg::<(), ()>::SendNft {
+                contract: receiver_contract.to_string(),
+                token_id: "test".to_string(),
+                msg: to_binary(&InnerMsg::Succeed).unwrap(),
+            },
+            &[],
+        )
+        .unwrap();
+
+    println!("{:?}", response);
```
Art3miX: We can remove the print and assert the attributes are what we expect them to be.

I would also add a test for the failed option and for wrong msg, just in case for future changes.

Art3miX (commented):
Cute little contract.

P.s, need CI fix.

contracts/cw721-receiver/Cargo.toml
chipshort: I moved cw721-base. cw721 is needed because I reuse the `Cw721ReceiveMsg` in the contract's execute msg.

contracts/cw721-receiver/tests/multitest.rs
chipshort: Done 👍

chipshort:
I applied @Art3miX's feedback.
Should I create a `test-contracts` folder and move it there?

Art3miX (approved):
LGTM

taitruong:
> I applied @Art3miX's feedback. Should I create a `test-contracts` folder and move it there?

Just rename it to `cw721-receiver-tester`, no extra folder needed imo. Once done we can merge it then.

chipshort:
Can someone merge? I don't have write access here

JakeHartnell (approved):
Amazing! This will be really useful.

PR #146, opened 2023-12-12
new features: setting and removing withdraw addresses by owner, allowing withdrawing funds by any to given address

taitruong:
New feature allowing cw721 withdrawing funds. This is e.g. needed for fee share module: https://docs.junonetwork.io/developer-guides/juno-modules/feeshare#registering-factory-contracts

Note: fee share is just an example. In general it allows cw721 holding funds and being able to withdraw. Like royalties and mint funds may be stored here.
also withdraw address is optional - so owner can opt to set a withdrawal address or not.

Added these new messages to cw721-base:

ExecuteMsg

```rs
    /// Sets address to send withdrawn fees to. Only owner can call this.
    SetWithdrawAddress { address: String },
    /// Removes the withdraw address, so fees are sent to the contract. Only owner can call this.
    RemoveWithdrawAddress {},
    /// Withdraw from the contract to the given address. Anyone can call this,
    /// which is okay since withdraw address has been set by owner.
    WithdrawFunds { amount: Coin },
```

QueryMsg:

```rs
    #[returns(String)]
    GetWithdrawAddress {},
```

@shanev @JakeHartnell

### DIFF
--- a/contracts/cw721-base/src/contract_tests.rs
+++ b/contracts/cw721-base/src/contract_tests.rs
@@ -2,7 +2,7 @@
 use cosmwasm_std::testing::{mock_dependencies, mock_env, mock_info};
 
 use cosmwasm_std::{
-    from_json, to_json_binary, Addr, CosmosMsg, DepsMut, Empty, Response, StdError, WasmMsg,
+    from_json, to_json_binary, Addr, Coin, CosmosMsg, DepsMut, Empty, Response, StdError, WasmMsg,
 };
 
 use cw721::{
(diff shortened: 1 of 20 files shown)


taitruong (commented):
All the key logic is here

contracts/cw721-base/src/execute.rs
```diff
@@ -127,6 +141,60 @@ where
         let ownership = cw_ownable::update_ownership(deps, &env.block, &info.sender, action)?;
         Ok(Response::new().add_attributes(ownership.into_attributes()))
     }
+
+    pub fn set_withdraw_address(
```
taitruong, line 145: All the key logic is here

contracts/cw721-base/src/contract_tests.rs
```diff
@@ -837,6 +845,86 @@ fn approving_all_revoking_all() {
     }
 }
 
+#[test]
+fn test_set_withdraw_address() {
```
taitruong, line 849: all tests are here

Issue #151, opened 2024-01-10
make minter for instantiate msg optional


taitruong:
Rather prefer having `minter: Option<String>`:
https://github.com/CosmWasm/cw-nfts/blob/177a993dfb5a1a3164be1baf274f43b1ca53da53/contracts/cw721-base/src/msg.rs#L8-L18

In case of none, `info.sender` should be minter. Makes it easier for other contracts (e.g. minter) re-using this msg.

Issue #150, opened 2024-01-10
release 0.19.0


taitruong:
Imo, we should publish a new release. The following have been added:

- #140: cw721-expiration: contract with expiration for NFTs
- #141: update libs
- #149: withdrawal address: contracts may have funds and send to optional withdraw address
- #147: MAX_LIMIT increased from 100 to 1000

@shanev @JakeHartnell

MbBrainz:
Is there a way I can help to accelerate this?

taitruong:
> Is there a way I can help to accelerate this?

It's in progress: #156

taitruong:
Updates:

- #156 merged - major v19 release
  - new creator ownership(!), along with existing minter ownership
  - optional(!) collection extension with royalty info (aka onchain collection metadata)
  - updatable NFT Info, including extension (aka onchain NFT metadata)
- #173 merged - minor adjustments like migrate msgs
- #168 includes new aggregated `GetConfig` query (includes num tokens, collection info, contract info, minter and creator)

Also it contains migration logic for upgrading cw721 v15 - v18 to latest v19 version.

0xphilipp:
Hey, just a heads up that the update to 0.19 creates much heavier contract sizes. 

Top one is a regular collection contract, while the bottom one is an advanced custom contract, but both increased ~400kb only by switching from 0.18 to 0.19

![image](https://github.com/user-attachments/assets/b426990a-8a71-4f94-b33a-e46223f3bd4b)

taitruong:
Yes, we r aware of this. There r plans for lite contract

Kiruse:
Is it intended that v0.19 is not on crates.io yet? If I am interpreting this screencap correctly, it should be published already:

![Image](https://github.com/user-attachments/assets/73c7aabc-1217-4e88-8623-7903b5d566bd)

PR #152, opened 2024-01-10
#151 make minter optional

taitruong:
PR for #151

Rather prefer having `minter: Option<String>`:
https://github.com/CosmWasm/cw-nfts/blob/177a993dfb5a1a3164be1baf274f43b1ca53da53/contracts/cw721-base/src/msg.rs#L8-L18

In case of none, `info.sender` should be minter. Makes it easier for other contracts (e.g. minter) re-using this msg.

In most cases creator of contract is also minter. Rn creator (=sender) instantiates contract AND need to provides its own address in `minter` prop. You can see this in unit tests:

Before: duplicate, since creator is in `info.sender` and `minter.prop`

```rust
        let info = mock_info(CREATOR, &[]);
        let init_msg = InstantiateMsg {
            name: "SpaceShips".to_string(),
            symbol: "SPACE".to_string(),
            minter: CREATOR.to_string(),
            withdraw_address: None,
        };
        entry::instantiate(deps.as_mut(), mock_env(), info.clone(), init_msg).unwrap();
```

After: no need to provider `minter` prop, since creator is sender:

```rust
        let info = mock_info(CREATOR, &[]);
        let init_msg = InstantiateMsg {
            name: "SpaceShips".to_string(),
            symbol: "SPACE".to_string(),
            minter: None,
            withdraw_address: None,
        };
        entry::instantiate(deps.as_mut(), mock_env(), info.clone(), init_msg).unwrap();
```

Instantiation is like this then:

like before, in case caller should NOT be minter/owner (edge case):
```json
{"name": "foo", "symbol": "bar", "minter": "other_than_caller"}
```

for creator (standard, 99.99% case):
```json
{"name": "foo", "symbol": "bar"}
```

@shanev @JakeHartnell

### DIFF
--- a/contracts/cw2981-royalties/src/lib.rs
+++ b/contracts/cw2981-royalties/src/lib.rs
@@ -136,7 +136,7 @@ mod tests {
         let init_msg = InstantiateMsg {
             name: "SpaceShips".to_string(),
             symbol: "SPACE".to_string(),
-            minter: CREATOR.to_string(),
+            minter: None,
             withdraw_address: None,
         };
         entry::instantiate(deps.as_mut(), mock_env(), info.clone(), init_msg).unwrap();
@@ -170,7 +170,7 @@ mod tests {
         let init_msg = InstantiateMsg {
             name: "SpaceShips".to_string(),
             symbol: "SPACE".to_string(),
-            minter: CREATOR.to_string(),
+            minter: None,
             withdraw_address: None,
         };
         entry::instantiate(deps.as_mut(), mock_env(), info.clone(), init_msg).unwrap();
@@ -201,7 +201,7 @@ mod tests {
         let init_msg = InstantiateMsg {
             name: "SpaceShips".to_string(),
             symbol: "SPACE".to_string(),
-            minter: CREATOR.to_string(),
+            minter: None,
             withdraw_address: None,
         };
         entry::instantiate(deps.as_mut(), mock_env(), info.clone(), init_msg).unwrap();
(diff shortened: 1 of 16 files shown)


taitruong:
You can also see from adjusted tests - it looks more convenient and natural.

contracts/cw2981-royalties/src/lib.rs
```diff
@@ -136,7 +136,7 @@ mod tests {
         let init_msg = InstantiateMsg {
             name: "SpaceShips".to_string(),
             symbol: "SPACE".to_string(),
-            minter: CREATOR.to_string(),
+            minter: None,
```
taitruong, line 139: If you scroll up, in line 135, you can see CREATOR is the sender

taitruong:
Reason, why I came to this conclusion is this here: https://github.com/public-awesome/launchpad/blob/main/contracts/collections/sg721-base/src/contract.rs#L53-L58

```rust
        // cw721 instantiation
        let info = CW721ContractInfoResponse {
            name: msg.name,
            symbol: msg.symbol,
        };
        self.parent.contract_info.save(deps.storage, &info)?;
        cw_ownable::initialize_owner(deps.storage, deps.api, Some(&msg.minter))?;
```

All this logic can completely be removed, by calling `parent.instantiate`. Also these 3 props can be replaced by `cw721_base::msg::InstantiateMsg`: https://github.com/public-awesome/launchpad/blob/main/packages/sg721/src/lib.rs#L118-L120

```rust
pub struct InstantiateMsg {
    pub name: String,
    pub symbol: String,
    pub minter: String,
    pub collection_info: CollectionInfo<RoyaltyInfoResponse>,
}
```

@jhernandezb

JakeHartnell (approved):
I'm fine with this, curious what others thinks, but agree it would be slightly nicer.

taitruong:
@shanev need one more approval, then we can merge and release 0.19.0 (#150)

shanev (approved):
Good change :)

PR #160, opened 2024-04-17
feat: upgrade to cosmwasm-std 2.0

PFC-developer:
This PR attempts to bring cw-nfts the latest cosmwasm-std version.

there are some breaking changes.

1. migrating from v16 NFTs is not supported
2. relies on a un-merged PR for cw-ownable (https://github.com/larry0x/cw-plus-plus/pull/22)
3. changes some lifetimes in storing key names from 'a to 'static
4. implementing a cw-util function parse_instantiate_reply_data which is no longer in the core package (and uses a deprecated 'data' field)

the bulk of the work is in the testing modules due to the ordering inside the tests (due to it requiring well from bech32 addresses, and the use of deps.api.addr_make() 

Thanks!


PFC-developer:
I'm going to leave the PR here for the moment.
I'm not sure why the multi-tests aren't being used (I didn't modify anything here)
and upgrading to a more recent cosmwasm-check requires a rust upgrade, and if you have a reason to stay on 1.65 or not.

contracts/cw2981-royalties/Cargo.toml
```diff
@@ -14,7 +14,7 @@ crate-type = ["cdylib", "rlib"]
 
 [features]
 # for more explicit tests, cargo test --features=backtraces
-backtraces = ["cosmwasm-std/backtraces"]
+# backtraces = ["cosmwasm-std/backtraces"]
```
webmaster128: You can remove those two lines entirely. cosmwasm-std/backtraces does not exist anymore in 2.0

> Also remove any uses of the `backtraces` feature. You can use a `RUST_BACKTRACE=1` env variable for this now.

contracts/cw2981-royalties/src/lib.rs
```diff
@@ -131,8 +131,9 @@ mod tests {
     fn use_metadata_extension() {
         let mut deps = mock_dependencies();
         let contract = Cw2981Contract::default();
+        let creator = deps.api.addr_make(CREATOR).to_string();
```
webmaster128: I'd always store those as owned `Addr`s, which is the most powerful type. You can then get a &str with `creator.as_str()`

contracts/cw2981-royalties/src/query.rs
```diff
@@ -19,7 +19,8 @@ pub fn query_royalties_info(
         },
         None => Decimal::percent(0),
     };
-    let royalty_from_sale_price = sale_price * royalty_percentage;
+    let royalty_from_sale_price =
+        Decimal::from_ratio(sale_price, Uint128::one()) * royalty_percentage;
```
webmaster128: `sale_price.mul_floor(royalty_percentage)` (or `mul_ceil`) is the new way to do Uint128 * Decimal = Uint128

contracts/cw721-base/src/query.rs
```diff
@@ -128,7 +128,7 @@ where
         let token = self.tokens.load(deps.storage, &token_id)?;
 
         // token owner has absolute approval
-        if token.owner == spender {
+        if token.owner.to_string() == spender {
```
webmaster128: This is a case sensitive address check. But there could be the same address on both sides. I would validate `spender` first and then perform the equality on two `Addr`s instead of `String`s

contracts/cw2981-royalties/src/lib.rs
PFC-developer: removed the to_strings where I could.

PFC-developer:
thanks for the review @webmaster128 .. I hope I got all the addr_makes's

noahsaso:
bump, we're blocked on this over at DAO DAO. is there anything i can do to push this forward?

taitruong:
@PFC-developer could u pls resolve conflicts? Regarding linter u dont need to worry. @jhernandezb will have a look at it.

PFC-developer:
THIS IS A WIP.
2 major issues 
1. cw_ownable relies on a unmerged PR. (OwnershipStore?)
2. something changed in the contract wrapper and the compile is failing.

I'm sure there are other issues, but it's not compiling due to these (and hence untested)

if you want to do start from scratch, I suggest you first remove the deprecated functions and use api_addr_make in pure v019, and then try to move things across to cosmwasm2

PR #156, opened 2024-02-27
Major release v0.19: add `CollectionInfo`, `RoyaltyInfo`, updatable NFTs for creator, etc.

taitruong:
# Major refactoring for upcoming v0.19 release.

TL;DR: v0.19 release is a major refactoring, introducing distinctions between Creator (manages collection metadata like royalties) and Minter (controls minting), and extends `UpdateNftInfo` with extension for onchain metadata. We've also added a new `CollectionInfo` structure within the cw721 package, moving all core logic for easier access and implementation across contracts, thus streamlining contract architecture. This update simplifies the ecosystem by removing the need for separate cw721-metadata-onchain and cw2981-royalty contracts and clearly defining roles and functionalities.

# Distinction between `Creator` and `Minter`

In previous release `cw721` is only aware of minter (aka owner/minter ownership). Here we added `Creator`. Creator controls collection info (like royalties) whereas minter controls minting.

# NEW utility: `UpdateNftInfo` and `UpdateCollectionInfo` msg

By adding a new `CollectionInfoExtension` store as an optional extension for `CollectionInfo` (name and symbol), there is also logic introduced here for updating collection metadata.

The creator is now able to do both: updating collection and NFT Info!

# NEW `CollectionMedata` in `cw721` package

```rs
#[cw_serde]
pub struct CollectionMetadata<TCollectionMetadataExtension> {
    pub name: String,
    pub symbol: String,
    pub extension: TCollectionMetadataExtension,
    pub updated_at: Timestamp,
}

#[cw_serde]
pub struct CollectionMetadataExtension<TRoyaltyInfo> {
    pub description: String,
    pub image: String,
    pub external_link: Option<String>,
    pub explicit_content: Option<bool>,
    pub start_trading_time: Option<Timestamp>,
    pub royalty_info: Option<TRoyaltyInfo>,
}

#[cw_serde]
pub struct RoyaltyInfo {
    pub payment_address: Addr,
    pub share: Decimal,
}
```

For making it available to all existing contracts, I had to move all core logic into `cw721` package. As a result, outcome is the following:

- there are now default implementations for the `Cw721Execute` and `Cw721Query` traits
- all structs like CollectionMetadata, NftMetadata, RoyaltyInfo, NftInfo, OwnerOfResponse, ApprovalResponse, ... moved to package

[...]

PR #161, opened 2024-04-25
Move to cw721 package

taitruong:
This PR only moves code from contracts to cw721 package - as a preparation for PR #156 (`Major release v0.19: add CollectionInfo, RoyaltyInfo, updatable NFTs for creator, etc.`).

@jhernandezb, @JakeHartnell, all checks passed - except lint checker fails with `use of deprecated variant ...`

I have marked deprecated usages with #[allow(deprecated)], but linter still complaints.
Did the same in launchpad PR, but here it is all fine:
https://github.com/public-awesome/launchpad/pull/668/


contracts/cw2981-royalties/src/msg.rs
```diff
@@ -24,16 +34,179 @@ pub enum Cw2981QueryMsg {
...
-        Cw2981QueryMsg::CheckRoyalties {}
+// impl Default for Cw2981QueryMsg {
+//     fn default() -> Self {
+//         Cw2981QueryMsg::CheckRoyalties {}
+//     }
+// }
+
+// impl CustomMsg for Cw2981QueryMsg {}
+
+impl From<QueryMsg> for Cw721QueryMsg<Extension> {
```
humanalgorithm, line 135: Is a goal of this PR to enable functionality of cw2981? We are adding all the messages here.

contracts/cw721-expiration/src/execute.rs
```diff
@@ -107,176 +123,72 @@ impl<'a> Cw721ExpirationContract<'a> {
...
-        self.assert_valid_nft(deps.as_ref(), &env, &token_id)?;
+    ) -> Result<Response<TCustomResponseMessage>, ContractError> {
+        self.assert_nft_expired(deps.as_ref(), &env, token_id.as_str())?;
         Ok(self
             .base_contract
             .revoke(deps, env, info, spender, token_id)?)
     }
 
-    fn approve_all(
+    pub fn transfer_nft_include_nft_expired(
```
humanalgorithm, line 155: Why are we changing the name here? What's the difference between approve_all and transfer nft include nft expired?

contracts/cw721-expiration/src/msg.rs
```diff
@@ -23,106 +23,123 @@ pub struct InstantiateMsg {
...
+    // -------- below adds `include_expired_nft` prop to cw721/src/msg.rs --------
     /// Return the owner of the given token, error if token does not exist
-    #[returns(cw721::OwnerOfResponse)]
+    #[returns(cw721::msg::OwnerOfResponse)]
     OwnerOf {
         token_id: String,
         /// unset or false will filter out expired approvals, you must set to true to see them
         include_expired: Option<bool>,
         /// unset or false will filter out expired nfts, you must set to true to see them
-        include_invalid: Option<bool>,
```
humanalgorithm, line 37: What is difference between "invalid" and "expired" ?

contracts/cw721-expiration/src/query.rs
```diff
@@ -1,296 +1,347 @@
...
-        include_expired: bool,
-        start_after: Option<String>,
-        limit: Option<u32>,
-    ) -> StdResult<OperatorsResponse> {
-        self.base_contract
-            .operators(deps, env, owner, include_expired, start_after, limit)
-    }
-
-    pub fn approval(
+    pub fn query_approval_include_expired_nft(
```
humanalgorithm, line 210: How come we add "query" in the name? Before they just say "approval" so I'm not sure if naming convention necessitates that we put "query" in the name or not?

contracts/cw721-expiration/src/query.rs
```diff
@@ -1,296 +1,347 @@
...
+            self.assert_nft_expired(deps, &env, token_id.as_str())?;
         }
         Ok(self
             .base_contract
-            .all_nft_info(deps, env, token_id, include_expired)?)
+            .query_all_nft_info(deps, env, token_id, include_expired_approval)?)
+    }
+
+    // --- helpers ---
+    pub fn is_nft_expired(&self, deps: Deps, env: &Env, token_id: &str) -> StdResult<bool> {
```
humanalgorithm, line 316: It seems like this "expired nft" is the bulk of the PR. What exactly is an "expired nft?"

humanalgorithm:
Can you better explain why we have this giant app centered around cw721 expiration? Shouldn't "expiration" kind of just be an option handed to a regular cw721 rather than a giant app?

humanalgorithm (approved):

contracts/cw2981-royalties/src/msg.rs
taitruong: as discussed, `cw2981` will be removed in next PR #156
Here it just got refactored. Before Cw721QueryMsg was wrapped inside Cw721's QueryMsg. So here it explicitely follows same struct as for cw721. But yeah - will be removed anyways. this contract is useless and not production ready, imo

contracts/cw721-expiration/src/execute.rs
taitruong: `transfer_nft_include_nft_expired` has nothing to do with `approve_all`. Diff just shows previous line has now different code.

Also here: most logic has been moved to package. Like `approve_all`. This `cw721-expiration` has some specific logic, where expiration needs to be considered. Like for `ExecuteMsg::TransferNft`, expiration needs to be considered in contract, where as other msgs (e.g. `ApproveAll`) is independent from expiration - hence handled in package.

contracts/cw721-expiration/src/msg.rs
taitruong: Prop `include_expired` is in general for cw721-base and covers expiration of approvals. Now cw721-expiration includes expiration of NFTs. Term `invalid` wasnt precised, so it got renamed `include_invalid` -> `include_expired_nft`.

Please note for cw721-expiration: I created this new contract for a specific use case (like issuing 12months subscription using NFT expiring after 12 months).

contracts/cw721-expiration/src/query.rs
taitruong: again, like above diff here, just show it propery. all general cw721 queries are moved into package. All contract specific logic are kept here.

contracts/cw721-expiration/src/query.rs
taitruong: Example: you provided subscriptions, like a user can subscribe a domain/node/any service for 12 months. Now you user can buy this service, by buying an NFT which expires after 12 months. So creator can create a platform, where user can approve/have access by this NFT, once it got expired, subscription ends.

taitruong:
> Can you better explain why we have this giant app centered around cw721 expiration? Shouldn't "expiration" kind of just be an option handed to a regular cw721 rather than a giant app?

cw721-expiration is specific - and should not be part for cw721-base. Besides contract is already part of previous version. This PR here just did some refactorings.

contracts/cw2981-royalties/src/msg.rs
shanev: We cannot remove cw2981 as its being used by other chains such as Sei. Ideally any changes are backwards compatible.

packages/cw721/src/execute.rs
```diff
@@ -0,0 +1,678 @@
...
+    fn instantiate(
+        &self,
+        deps: DepsMut,
+        _env: Env,
+        info: MessageInfo,
+        msg: Cw721InstantiateMsg,
+        contract_name: &str,
+        contract_version: &str,
+    ) -> Result<Response<TCustomResponseMessage>, Cw721ContractError> {
+        cw2::set_contract_version(deps.storage, contract_name, contract_version)?;
```
taitruong, line 40: @shanev look here. Like `Cw721Execute` in cw721 package. It is still an interface/trait. Only difference here is that it provides a default implementation like here in `instantiate()`. This way contract just can implement `Cw721Execute` and in case it custom logic is required, contract can implement its own logic - but in 99% of the cases it can just use default logic.

taitruong: We did the same for `cw-ics721`, by providing default logic - like many other projects: https://github.com/public-awesome/cw-ics721/blob/main/packages/ics721/src/execute.rs#L36-L48

Cargo.toml
```diff
@@ -16,10 +16,15 @@ cosmwasm-std    = "^1.2"
...
+cw721-017       = { git = "https://github.com/CosmWasm/cw-nfts", tag = "v0.17.0", package = "cw721" } # needed for testing legacy migration
+cw721-018       = { git = "https://github.com/CosmWasm/cw-nfts", tag = "v0.18.0", package = "cw721" } # needed for testing legacy migration
 cw721-base      = { version = "*", path = "./contracts/cw721-base" }
-cw721-base-016  = { version = "0.16.0", package = "cw721-base" }
+cw721-base-016  = { git = "https://github.com/CosmWasm/cw-nfts", tag = "v0.16.0", package = "cw721-base" } # needed for testing legacy migration
+cw721-base-017  = { git = "https://github.com/CosmWasm/cw-nfts", tag = "v0.17.0", package = "cw721-base" } # needed for testing legacy migration
+cw721-base-018  = { git = "https://github.com/CosmWasm/cw-nfts", tag = "v0.18.0", package = "cw721-base" } # needed for testing legacy migration
 cw-multi-test   = "^0.20"
-cw-ownable      = "^0.5"
+cw-ownable      = { git = "https://github.com/public-awesome/cw-plus-plus.git", branch = "multiple_ownership"} # TODO: switch to official https://github.com/larry0x/cw-plus-plus once merged
```
JakeHartnell: Where did this come from? If we're going to use a `git` dependency, can we at least tie it to a specific commit / rev.

contracts/cw2981-royalties/src/msg.rs
```diff
@@ -24,16 +34,179 @@ pub enum Cw2981QueryMsg {
...
-    fn default() -> Self {
-        Cw2981QueryMsg::CheckRoyalties {}
+// impl Default for Cw2981QueryMsg {
+//     fn default() -> Self {
+//         Cw2981QueryMsg::CheckRoyalties {}
+//     }
+// }
+
+// impl CustomMsg for Cw2981QueryMsg {}
+
```
JakeHartnell:
```suggestion
```

contracts/cw721-expiration/examples/schema.rs
```diff
@@ -1,10 +1,10 @@
 use cosmwasm_schema::write_api;
-use cw721_expiration::msg::{ExecuteMsg, InstantiateMsg, QueryMsg};
+use cosmwasm_std::Empty;
+use cw721_expiration::msg::{InstantiateMsg, QueryMsg};
 
 fn main() {
     write_api! {
         instantiate: InstantiateMsg,
-        execute: ExecuteMsg,
```
JakeHartnell, line 7: Nit, should include execute messages in schemas because `ts-codegen` needs them for frontend tooling.

packages/cw721/examples/schema.rs
JakeHartnell, line 1: Not sure if we need a schema for the package.

packages/cw721/schema/all_nft_info_response.json
JakeHartnell, line 1: Also don't need all of these JSON files for each response type commited. I vote we get rid of `schema.rs` and these schema files for the `cw721` package.

contracts/cw721-expiration/examples/schema.rs
taitruong: Thx! Missed that. Fixed and added.

contracts/cw2981-royalties/src/msg.rs
taitruong: This was some test code from me, I removed it.

Cargo.toml
taitruong: Fixed. now using a revision.
`cw-ownable` is a singleton, made a fork so it can used multiple times, Shane and Jorge also says this makes sense. Larry said he's fine with it and wanted to merge (which he hasnt done yet): https://github.com/larry0x/cw-plus-plus/pull/21

packages/cw721/examples/schema.rs
taitruong: Quick note: there was schemas before the change. it looks like ts-codegen checks them for generating client code.

shanev (approved):

JakeHartnell (approved):

PR #166, opened 2024-07-03
Fix optimizer call in build.sh

webmaster128:
- Remove unneeded `--platform` arg (optimizer does not have multi-arch support in order to get better reproducibility. ARM images have a different name)
- Fix location of target folder to make build cache work again
- Use cosmwasm/optimizer image as cosmwasm/workspace-optimizer is a deprecated alias for that


JakeHartnell (approved):

jhernandezb (approved):

taitruong:
@jhernandezb we can merge this, but before we need to get CI running again.

PR #179, opened 2025-02-12
Cosmwasm 2.0+

deniszagumennov:
Based on https://github.com/public-awesome/cw-nfts/pull/172 Pull Request, another attempt to update CosmWasm in the contracts


packages/cw721/src/execute.rs
```diff
@@ -103,7 +106,7 @@ pub fn initialize_creator(
     api: &dyn Api,
     creator: Option<&str>,
 ) -> StdResult<Ownership<Addr>> {
-    CREATOR.initialize_owner(storage, api, creator)
```
jhernandezb: not really sure if we should get rid of the CREATOR alias? in v0.18 it was introduced along with MINTER so that we can differentiate more clearly different behavior

contracts/cw2981-royalties/src/query.rs
```diff
@@ -30,7 +30,7 @@ pub fn query_royalties_info(
         },
         None => Decimal::percent(0),
     };
-    let royalty_from_sale_price = sale_price * royalty_percentage;
+    let royalty_from_sale_price = sale_price.mul_floor(royalty_percentage);
```
jhernandezb, line 33: should we do `mul_ceil` instead ?  it will be losing a bit on flooring the result

Cargo.toml
```diff
@@ -25,11 +26,12 @@ cw721-base-016  = { git = "https://github.com/CosmWasm/cw-nfts", tag = "v0.16.0"
...
 cw721-base-017  = { git = "https://github.com/CosmWasm/cw-nfts", tag = "v0.17.0", package = "cw721-base" } # needed for testing legacy migration
 cw721-base-018  = { git = "https://github.com/CosmWasm/cw-nfts", tag = "v0.18.0", package = "cw721-base" } # needed for testing legacy migration
-cw-multi-test   = { version = "^0.20", features = ["cosmwasm_1_2"] }
-cw-ownable      = { git = "https://github.com/public-awesome/cw-plus-plus.git", rev = "28c1a09bfc6b4f1942fefe3eb0b50faf9d3b1523"} # TODO: switch to official https://github.com/larry0x/cw-plus-plus once merged
-cw-paginate-storage = { version = "^2.4", git = "https://github.com/DA0-DA0/dao-contracts.git" }
-cw-storage-plus = "^1.1"
-cw-utils        = "^1.0"
+cw-multi-test   = { version = "^2.1", features = ["cosmwasm_2_0"] }
+# TODO: switch to official https://github.com/larry0x/cw-plus-plus
+# requires merge of https://github.com/larry0x/cw-plus-plus/pull/25
```
jhernandezb:
```suggestion
```

packages/cw721/src/state.rs
```diff
@@ -49,14 +48,14 @@ pub struct Cw721Config<
     TNftExtension: Cw721State,
 {
     /// Note: replaces deprecated/legacy key "nft_info"!
-    pub collection_info: Item<'a, CollectionInfo>,
-    pub collection_extension: Map<'a, String, Attribute>,
-    pub num_tokens: Item<'a, u64>,
+    pub collection_info: Item<CollectionInfo>,
+    pub collection_extension: Map<String, Attribute>,
+    pub token_count: Item<u64>,
```
jhernandezb: this will probably break some stuff rather leave the same name here if there is no real reason to change it?

packages/cw721/src/msg.rs
```diff
@@ -464,29 +464,41 @@ impl StateFactory<CollectionExtension<RoyaltyInfo>>
...
+        let sender = info.map(|i| &i.sender);
         // start trading time can only be updated by minter
-        let minter_initialized = MINTER.item.may_load(deps.storage)?;
+        let is_minter_or_uninitialized = match sender
+            .map(|addr| MINTER.is_owner(deps.storage, addr))
+            .transpose()
+        {
+            // if Some(bool), bool represents the `is_owner` check result.
+            // if None, `info` was none and the check is skipped.
+            Ok(is_minter) => is_minter.unwrap_or(true),
```
humanalgorithm: I feel like this variable should be named something like  is_minter_owner ?

packages/cw721/src/msg.rs
```diff
@@ -464,29 +464,41 @@ impl StateFactory<CollectionExtension<RoyaltyInfo>>
...
+            && MINTER.item.exists(deps.storage)
+            && !is_minter_or_uninitialized
         {
             return Err(Cw721ContractError::NotMinter {});
         }
         // all other props collection extension can only be updated by the creator
-        let creator_initialized = CREATOR.item.may_load(deps.storage)?;
+        let is_creator_or_uninitialized =
+            match sender.map(|addr| is_owner(deps.storage, addr)).transpose() {
+                // if Some(bool), bool represents the `is_owner` check result.
```
humanalgorithm: Generally I'm wary of comments explaining code, is there a way to name the variables so its self documenting?

packages/cw721/src/state.rs
```diff
@@ -82,20 +81,20 @@ where
...
-        collection_info_key: &'a str,
-        collection_info_extension_key: &'a str,
-        num_tokens_key: &'a str,
-        operator_key: &'a str,
-        nft_info_key: &'a str,
-        nft_info_owner_key: &'a str,
-        withdraw_address_key: &'a str,
+        collection_info_key: &'static str,
+        collection_info_extension_key: &'static str,
+        token_count_key: &'static str,
```
humanalgorithm: What' sthe difference between "num_tokens_key" and "token_count_key" ?

shanev (commented):
LGTM as long as the comments by @jhernandezb and @humanalgorithm are addressed.

contracts/cw2981-royalties/src/query.rs
deniszagumennov: it would break this test: 

https://github.com/public-awesome/cw-nfts/blob/a5abe476c1028b2563f995adab184b86e3fc03ff/contracts/cw2981-royalties/src/lib.rs#L335 (in the old branch) 

and https://github.com/hadronlabs-org/cw-nfts/blob/df9fb7a8cd0f3e27ba0f4a17649df64069d09ad4/contracts/cw2981-royalties/src/lib.rs#L335 (in the new branch)

Cargo.toml
deniszagumennov: Removed this

packages/cw721/src/state.rs
deniszagumennov: I undid this change

packages/cw721/src/msg.rs
deniszagumennov: Renamed it

packages/cw721/src/state.rs
deniszagumennov: I undid this change

packages/cw721/src/msg.rs
deniszagumennov: I tried, please check it out

packages/cw721/src/execute.rs
deniszagumennov: they hid the `OWNERSHIP` field that `CREATOR` was based on (https://github.com/larry0x/cw-plus-plus/blob/2e33b2bdff5990e9126bbd87e1584ce0827cd2c8/packages/ownable/src/lib.rs#L261), but I was able to restore it, please see if it's any better

contracts/cw2981-royalties/src/query.rs
jhernandezb: ah makes sense I think maybe we should just follow up on a new issue thanks!

jhernandezb (approved):

packages/cw721/src/execute.rs
jhernandezb: lgtm

shanev (approved):
LGTM

shanev:
Thank you @deniszagumennov !!

PR #186, opened 2026-02-15
feature gate cw721_v016 for edge case conflicts with later cosmwasm-std versions

hard-nett:
feature gate cw721_v016 for edge case conflicts with later cosmwasm-std version


hard-nett:
@jhernandezb

hard-nett:
closing in favor of #183

